Issue #6, opened 2015-08-12
setup and document docker


justinlittman:
To only include rabbitmq service.

dchud:
New repo to be "sfm-docker".

dchud:
Also: investigate adding other GW Libs staff to have sufficient admin rights on docker hub.

kerchner:
Succcessor to #8

justinlittman:
@kerchner added as an owner of [gwul](https://hub.docker.com/u/gwul/) organization.

justinlittman:
See https://github.com/gwu-libraries/sfm-docker/commit/d1b5ea22cf059695288dec632fca1c79c97dc909 for setup and documenting docker.

Will add documentation once docs directory is setup.

rajatvij:
Pull request accepted. hence closing the ticket.

PR #7, opened 2015-08-13
Setup and document docker

justinlittman:
refs #6


justinlittman:
Determined that this pull request should not go against this repo, so closing.

PR #19, opened 2015-08-28
T18 initial django layout

rajatvij:


justinlittman:
I'd suggest the SFM django app belongs in a separate repo, not the main repo.

PR #20, opened 2015-08-31
T12 initial login

rajatvij:


dchud:
Just to confirm, does this now need to be synced from master?  I'll reassign back to @rajatvij if so.

rajatvij:
Let me check once again as i was having some trouble migrating data once i merged this with master on my local.

rajatvij:
Updated settings file. This can now be merged with master. Reassigning back to Dan C

rajatvij:
https://guides.github.com/features/mastering-markdown/

dchud:
@rajatvij, I've tested it out, and I think we should discuss it together before merging.  I see that the allauth pieces go in, and the login panel pops up, but we should probably add a default home page and some other bits to it first.  Also, I know I've seen a lot of styling and such added in your dev instance, and I don't see any of that here - I'd like to discuss how your work on this ticket/pr relates to #14 / #26 before proceeding.

rajatvij:
Sure. I scraped all the styling in these branches as those can be added in the baseline UI theme once we start working on it. And yes i wished to discuss before building a home screen so i left that out.

PR #23, opened 2015-09-02
T10 initial data models

dchud:


rajatvij:
@dchud In SeedSet Model i can see both date_added and date_started field. Should this be removed? And are we leaving the Model for Issues for later?

dchud:
Let's leave Issues for later, yes.

I meant for "started" to mean when collection of the seeds began, vs
"added" meaning when the set itself was first established. This might allow
us to create a set today for collection later on, and to record the
difference. But you're right to point out that it isn't that well thought
out!

Let's leave it in as a placeholder to resolve in a following release.


On Sep 2, 2015, at 3:36 PM, RAJAT <notifications@github.com> wrote:

@dchud <https://github.com/dchud> In SeedSet Model i can see both
date_added and date_started field. Should this be removed? And are we
leaving the Model for Issues for later?

—
Reply to this email directly or view it on GitHub
<https://github.com/gwu-libraries/sfm-ui/pull/23#issuecomment-137222532>.

PR #24, opened 2015-09-08
refs #6. Documentation on installation and rabbitmq.

justinlittman:


rajatvij:
@justinlittman Although the steps are pretty straightforward, but can we add channel declaration steps in RabbitMQ document as well, as that would make it so that we wont need to refer to RabbitMQ document at all if we need to just declare some exchange and get something working.

justinlittman:
@rajatvij Added an example showing channel declaration steps and binding.

PR #26, opened 2015-09-08
T14 add seed seed sets

rajatvij:
This can be merged with #13 as this covers both #13 and #14


dchud:
Note the two conflicts in settings/base.py and requirements.txt - please update to fix, they are both fairly small.

rajatvij:
closing because of #32

Issue #21, opened 2015-09-02
create docker containers for sfm ui


justinlittman:

justinlittman:
Suggest using the approach for containerizing a django app using for original social feed manager.

justinlittman:
@rajatvij - I've updated the docker-compose in https://github.com/gwu-libraries/sfm-docker/tree/sfm_t21-docker_for_ui to instantiate the containers.  When I do, I get:
```
Waiting for db
python: can't open file '/opt/sfm-setup/appdeps.py': [Errno 2] No such file or directory
Updating requirements
Requirement already satisfied (use --upgrade to upgrade): django<=1.9,>=1.8 in /usr/local/lib/python2.7/dist-packages (from -r /opt/sfm-ui/requirements/requirements.txt (line 1))
Requirement already satisfied (use --upgrade to upgrade): pytz in /usr/local/lib/python2.7/dist-packages (from -r /opt/sfm-ui/requirements/requirements.txt (line 2))
Requirement already satisfied (use --upgrade to upgrade): psycopg2 in /usr/local/lib/python2.7/dist-packages (from -r /opt/sfm-ui/requirements/requirements.txt (line 3))
Requirement already satisfied (use --upgrade to upgrade): django-allauth in /usr/local/lib/python2.7/dist-packages (from -r /opt/sfm-ui/requirements/requirements.txt (line 4))
Requirement already satisfied (use --upgrade to upgrade): django-crispy-forms in /usr/local/lib/python2.7/dist-packages (from -r /opt/sfm-ui/requirements/requirements.txt (line 5))
Requirement already satisfied (use --upgrade to upgrade): requests-oauthlib>=0.3.0 in /usr/local/lib/python2.7/dist-packages (from django-allauth->-r /opt/sfm-ui/requirements/requirements.txt (line 4))
Requirement already satisfied (use --upgrade to upgrade): python-openid>=2.2.5 in /usr/local/lib/python2.7/dist-packages (from django-allauth->-r /opt/sfm-ui/requirements/requirements.txt (line 4))
Requirement already satisfied (use --upgrade to upgrade): requests>=1.0.3 in /usr/lib/python2.7/dist-packages (from django-allauth->-r /opt/sfm-ui/requirements/requirements.txt (line 4))
[...]

PR #30, opened 2015-09-23
For #21 - docker for sfm-ui

justinlittman:


justinlittman:
Notes for testing:
* At this point, only the dev version will work.  Master will work once this is merged to master.  Production will work once the 0.2.0 tag is created.
* To test, rename the example docker-compose to something else.  Also, change the volume to point to your local copy of sfm-ui and replace the image arguments with build arguments.  (The images won't work until we add them to docker hub.)

rajatvij:
@justinlittman I tried for the dev-app. But i am getting No such image: error while running for sfmdevapp when i run: 
docker-compose up -d 
even when i can see the image when i do 
docker images

It must be some configuration issue and i will look into it a bit more.

justinlittman:
Here's the docker-compose I use:

```
sfmdevdb:
    build: ./db
    environment:
        - POSTGRES_PASSWORD=gherD42#dl5
sfmdevapp:
    build: ./app-dev
    ports:
        - "8081:80"
    links:
        - sfmdevdb:db
    environment:
        - SFM_DEBUG=True
    volumes:
        - ~/Data/sfm3/sfm-ui:/opt/sfm-ui
```

PR #32, opened 2015-09-28
T14 add seed seedsets try2

rajatvij:


justinlittman:
The code looks clean.  However:
1. It would be really helpful to provide some notes for the tester.
2. Some sort of a basic home page is necessary so that a user can get to collections, seeds, and/or seedsets without reading urls.py.

rajatvij:
@justinlittman I have added the navigation links to collection, seed and seedset list for testing in the branch. Please check and let me know if you need any help.

justinlittman:
Navigation looks good to me.

justinlittman:
Question:  Is Credentials.token intended to contain json?  Or something else?  I want to make sure we can handle different credentials for different platforms.

justinlittman:
Here are my recommended changes to align/clarify vocabulary:
* SeedSet.platform --> harvest_type
* SeedSet.crawl_options --> harvest_options
* SeedSet.date_started --> start_date
* SeedSet.date_ended --> end_date
* Seed.platform_token --> seed_token
* Seed.platform_uid --> seed_uid

justinlittman:
@rajatvij, @dchud :  Throwing this out for consideration -- Do we want to remove fields/models that are speculative?  We can add them back in later when we're ready to make use of them and better understand how they would be used.

In particular:
* *.stats
* Collection.is_visible
* Harvest
* Media

dchud:
@justinlittman - re the date fields, I'd prefer that they be consistent throughout the models.  There are other date fields elsewhere, and I'm hesitant to say we should spend more cycles this iteration renaming everything.

Re your suggested Seed.seed_token and Seed.seed_uid, it seems redundant to say "seed" so often, as it's already a seed. :)

dchud:
@justinlittman re Credentials.token, it might be redundant relative to the stored credentials obtained via all_auth.

justinlittman:
I'm not suggesting changing all dates, just those 2.  Unlike all other dates, they refer to a future date rather than recording the date of something that has already happened.  By putting it in the past tense, the labels for those 2 dates is confusing -- you mean "start date" and "end date" not "date started" and "date ended".  (For the record, I was confused during the code review.)

Also, I've changed all of the dates in the messages to follow the date_XXXXed convention.

OK with dropping seed_ from token and uid.  More important is dropping platform_.

dchud:
:+1:

rajatvij:
@justinlittman I have updated the field names as per above comments. Can you please check and let me know if there is anything missing in my changes.

Issue #33, opened 2015-09-29
Formatting of sfm admin is broken


justinlittman:
Steps to reproduce:
1. Go to sfm admin (e.g., http://192.168.99.100:8081/admin/)
2. Log in (using sfmadmin/password).

Expected result:
Correctly rendered sfm admin home page.

Actual result:
![screen shot 2015-09-29 at 9 13 51 am](https://cloud.githubusercontent.com/assets/588335/10164946/73b2d9e6-668a-11e5-87dc-d8e8f08b309d.png)

justinlittman:
It's caused by a 403 on `/static/admin/css/base.css` and `/static/admin/css/dashboard.css`.

PR #34, opened 2015-09-29
For #33 - Fixes admin formatting

justinlittman:


rajatvij:
Seems good to me. This fixes #33

PR #39, opened 2015-10-14
T16 messaging tracer bullet

rajatvij:
Hi Justin, we can send a message tracer bullet by updating seeds. Please check and let me know if i need to construct the json message in a particular way or if this is fine for now.


kerchner:
:+1: !

rajatvij:
@justinlittman I have updated views to send messages on creation, updation and deletion of seeds and cleaned views.py a bit. 
Currently app is sending simple json messages to rabbit worker at the moment. I am trying to create a message as close as possible to messaging spec right now. So do let me know if that is required in this pull request or not.

justinlittman:
Here are my code review/testing comments:
1.  A rabbit container should be added to example.dev.docker-compose.yml since it is necessary to run the app.
2.  Please provide suggestions for testing.  This should include instructions on how to setup a test environment (or point to the appropriate instructions), a testing procedure (that is, "try this ..."), and any necessary sample values (e.g., you can use 131866249@N02 as a valid flickr uid).
3.  This PR contains extraneous commits.  I think we're going to have to do another one of those kill this PR and construct a clean branch and copy over the code.
4.  When I log into Site Admin, Collections, Credentials, Seed Sets, and Seeds section is called RabbitWorker.  Seems like something is wrong.
5.  In rabbit.py:  Why do you declare and bind to queue?  For the tracer bullet you should only be publishing to the exchange.
6.  I'd suggest that instead of tying the message submit to Seed*View.post(), just add a button on the seedset page for sending that seed set.
7.  When I try to add a seed, I get a `ConnectionClosed at /ui/seeds/create/` error.

Yes, sending a complete message is required for this ticket.  The harvest type, seeds, options, collection id, and credentials should come from the model; everything else can be jerry-rigged.

Let me know if you'd like to discuss any of this.

rajatvij:
@justinlittman I have resolved the merge conflicts and removed declaration and binding of queue from rabbit.py. 

Rabbit Worker was coming from verbose_name in rabbit.py, so i have corrected that too.

I will work on adding a button to send message instead of sending it via submit button and compose a complete message.

At the moment, this is sending message in following format.

By Updating a Seed
message = {"Action": "Update Seed", "Token": "tests", "SeedSet Object": "1", "UID": "tests"}
By Creating a Seed
message = {"Action": "Create Seed", "Token": "tests", "SeedSet Object": "1", "UID": "tests"}
By Deleting a Seed
message = {"Action": "Delete Seed", "ID": "1"}

Steps to reproduce:
1. Go to admin page and check if there is any group already present, if not create one.
2. Create a collection(either from admin or from UI).
3. Create a seedset(either from admin or from UI).
4. Create a seed from UI to generate a create seed message.
5. Update a seed from UI to generate an update seed message.
6. Delete a seed from UI to generate a delete seed message.

I am trying to compose a complete seed message according to Messaging spec document in wiki and will update on it once i am done with it. I will need some time though as i am not that familiar on how to fetch data referred from other models or tables in database, that's why i produced messages containing fields present in a single model for the moment.

I am not getting ConnectionClosed at /ui/seeds/create/ error, we can discuss this later on Monday.

rajatvij:
@justinlittman I have update the views to send tracer bullet only on Updating Seed Sets. Please ignore my above message's instructions as tracer bullet is only sent on updating seed set now.

Steps to reproduce:
1. Create a collection, if not present.
2. Create a seedset, if not present. [NOTE: I gave a json text in harvest option field in seed set here]
3. Create seeds and tie them to seedset.
4. To send tracer bullet, Update SeedSet. Tracer bullet will be sent on clicking submit.

Tracer bullet generated 

{"collection": [{"id": 1}], "options": "{ 'sizes': ['Thumbnail', 'Large', 'Original'] }", "seeds": [{"token": "justin.littman", "id": 35, "uid": "131866249@N02"}, {"token": "test seed", "id": 68, "uid": "test uid"}], "credentials": [{"platform": "test", "token": "test", "id": 1, "user": 1}], "type": "flickr", "id": 2}

There are few discrepancies while comparing with the JSON message as per the messaging spec document.
1. In credentials, it seems we are not tying key in secret in our model so i wasnt able to get those values.
2. I am not sure on how include path in collection section, so i havent included it yet.

I will try to include a button to send tracer bullet and we can discuss on the above discrepancies on Monday.

rajatvij:
@justinlittman I fixed the routing key value as per our discussion.
To have correct key value we need to provide correct information in platform field of credentials model and harvest_type field of our seed set.
In our case
we can use 'flickr' as platform in related credential and 'flickr_user' as harvest_type in seed_set.

Let me know if any more changes are required.

rajatvij:
@justinlittman I have updated views as per our discussion.
For now i am using a dirty way to remove /n/r characters as i wasn't able to do it using json library. I will try to do it later tomorrow. But i guess this should be able to resolve the issues you were having.
Let me know if i have to consider any other case apart from '/r/n' characters.

justinlittman:
This will give you:
```
"credentials": "{\"key\": \"abddfe6fb8bba36e8ef0278ec65dbbc8\",\"secret\": \"1642649c54cc3ebe\"}"
```
what you need is:
```
"credentials": {"key": "abddfe6fb8bba36e8ef0278ec65dbbc8","secret": "1642649c54cc3ebe"}
```

justinlittman:
@rajatvij, @kerchner  I think the last thing to do is to figure out how to clean up the commits so that they only contain this tickets commits.  My suggestion is to just create a new branch from master and manually apply the changes rather than git magic.  Suggestions?

rajatvij:
@justinlittman I can work on that, although there wont be very many changes from what this branch currently has as most of the older commits got overridden with later commit to this branch.

@kerchner Let me know if i should follow what Justin is proposing.

Issue #16, opened 2015-08-17
implement messaging tracer bullet


dchud:
Depends on #14.  Workers can just log to console when messages are received / passed further along.

justinlittman:
Depends on #22.

rajatvij:
I have been trying to create a json message out of forms on seed update page. By our use case i think the json messages should be generated when a seed, seed set or collection is created or updated. Theoretically using get_object function in view we should be able to get the values of the fields in view itself from where we can create and push json messages on queue. But i can't seem to get around it yet. So i would appreciate any help i can get on this. Or if there is any other approach that i should use.
I can always use javascript to get around this but i think finding a django way to do this would make our code more cleaner as compared to using javascript in this case.

justinlittman:
This is all only a temporary step until we add a scheduler.  I would
suggest just adding a "send message" button that triggers sending the
message for the seed set.  No need for anything more complicated.

kerchner:
Per discussion, the form should submit to the django view, and the view code should trigger the JSON message being sent to the queue - rather than sending the message from the Javascript on the client side.

justinlittman:
The tracer bullet consists of:
1. A user populating collection, seed, seed set, and credential information in the SFM UI for a Flickr user .  (Parts of this can be performed in Admin.)
2. Some action by the user (clicking a button, submitting a form) triggers constructing a Flickr user harvest start message and submitting to the queue.
3. The Flickr Harvester receives the message and performs the harvest.

Some additional details:
* For JSON fields in the model, the user can enter JSON into the text box.  This includes Credentials.token and SeedSet.harvest_options.
* Wherever necessary, parts of the harvest message can be hardcoded.
* The routing key for the message should be `harvest.start.flickr.flickr_user`.
* The harvest type is `flickr_user`.

Following is an example Flickr user harvest start message:
```
{
            "id": "test:1",
            "type": "flickr_user",
            "seeds": [
                {
                    "uid": "131866249@N02"
                }
            ],
            "options": {
                "sizes": ["Thumbnail", "Original"]
            },
            "credentials": {
                "key": "FLICKR_KEY",
                "secret": "FLICKR_SECRET"
            },
            "collection": {
                "id": "test_collection",
                "path": "/tmp/test_collection"

            }
}
```

PR #41, opened 2015-10-23
refs #16. Fire harvest start message on seed update.

justinlittman:


justinlittman:
@rajatvij I created this branch by cherry-picking commits from the original feature branch on top of a fresh branch created from master.  I then did a rebase -i to squash into a single commit.

Give it a try.

rajatvij:
@justinlittman we don't really need any changes in templates as those were what i used when i was trying to send tracer bullet using javascript. But now since we are not using that so i am removing those changes. I will test it and let you know if anything is amiss.

rajatvij:
@justinlittman this looks good to me

Issue #37, opened 2015-09-30
Add support for test/dev fixtures


justinlittman:
In development and CI environments, it would be helpful to have the ability to have fixtures preloaded.  At the very least, this should include a test user.

justinlittman:
This fixture creates a user called "sfmuser" with password "password".

To do:  Add this to documentation.

rajatvij:
Merged t37 branch hence closing the issue

PR #68, opened 2015-12-01
refs #37. Added fixtures to dev and master docker containers.

justinlittman:


justinlittman:
Steps to test (sfm-ui dev container):
1. Change your docker-compose.yml to build sfm-ui instead of using image.
2. Bring up sfm-ui with `docker-compose up -d`.
3. Log in as testuser (with password "password").

Issue #66, opened 2015-11-25
Change Timezone from GMT to EST in RabbitMQ


rajatvij:
Currently all messages in RabbitMQ are timestamped in GMT timezone which is different from our timezone. So to have a better clarity i suggest to use EST(GMT - 5hr) timezone or to include timezone along with time.

For eg, 2015-11-25T17:08:39.951739 EST instead of 2015-11-25T17:08:39.951739

kerchner:
Don't use EST specifically.  Get the timezone from the server.

justinlittman:
See https://github.com/docker/docker/issues/12084 and http://serverfault.com/questions/683605/docker-container-time-timezone-will-not-reflect-changes

rajatvij:
@justinlittman and @kerchner For this we have two ways to achieve this. 

1. We can either make changes in each Dockerfile like [this](http://tedwise.com/2015/05/02/setting-the-timezone-in-a-docker-image) 

2. Or at docker-compose level like [this](https://github.com/docker/compose/issues/1377) by mounting our localtime and timezone as volumes to the images. 

Personally i prefer the 2nd way as that takes care of the setting up timezone automatically and we don't need to specify timezone there and in first case we are only able to update local timezone, but not the universal timezone. And we would need to updated Dockerfile of each image which i am not even sure we are having. 

So please let me know which approach to follow and i will push the changes accordingly.

justinlittman:
I've also seen reference to setting the TZ env variable in the docker-compose (e.g, `TZ=America/Los_Angeles`).  Can you give this a try?

For 2, will the location of the volumes on a user's machine vary?  Our is it the same on all machines/OSes?

rajatvij:
For 2 we just need to mount these volumes to each docker container: 

volumes:
    - /etc/localtime:/etc/localtime:ro
    - /etc/timezone:/etc/timezone:ro

And for 1st it works even without setting a TZ env variable in Docker file. We just need to add following lines in Dockerfile, but this would only change the timezone of that single container and here we need to specify the timezone:

# Set the timezone.
RUN sudo echo "America/New_York" > /etc/timezone
RUN sudo dpkg-reconfigure -f noninteractive tzdata

justinlittman:
Can you try just setting a TZ environment variable in the docker-compose and see if it works?  (See, for example, the discussion in https://github.com/sameersbn/docker-gitlab/issues/113).

[...]

PR #102, opened 2015-12-20
refs #66. Updated example.docker-compose.yml to change timezone.

rajatvij:
In this we just have to update our docker-compose.yml and include following volumes in each container to have local time synced your machine's time:

volumes:
- /etc/localtime:/etc/localtime:ro
- /etc/timezone:/etc/timezone:ro

Let me know if we need to use the other approach we discussed here : https://github.com/gwu-libraries/sfm-ui/issues/66

I have updated example.dev.docker-compose.yml accordingly for reference.

### DIFF
--- a/docker/example.dev.docker-compose.yml
+++ b/docker/example.dev.docker-compose.yml
@@ -5,11 +5,13 @@ sfmdevrabbit:
         - "15672:15672"
     restart: always
     environment:
+        - TZ=EST
         - RABBITMQ_DEFAULT_USER=sfm_user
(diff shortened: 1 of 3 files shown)


kerchner:
@justinlittman will implement approach described in issue.

rajatvij:
@justinlittman Please check if all logs are getting correct timezone as specified in docker-compose file. Although i cant see any problem anywhere. But i think if we have set any logs to get time from Universal time instead of Local time of container, then we would need to change that.

Issue #91, opened 2015-12-17
Job improvements


justinlittman:
In support of #79.

Clean-up creating harvest messages and add unit tests.

justinlittman:
To do: Add travis configuration.

PR #103, opened 2015-12-21
T91 job improvements

justinlittman:
refs #91


justinlittman:
PR approved by @rajatvij

kerchner:
@justinlittman @rajatvij This PR broke storing the Harvest result - see https://github.com/gwu-libraries/sfm-ui/pull/103/files#diff-7ae2cd25da611053b5b37d8e7cb2f240R26.  sfm_ui_consumer.py had been using `id` as the seedset id to link the harvest status message with the SeedSet.

kerchner:
@justinlittman @rajatvij With that said, it looks like https://github.com/gwu-libraries/sfm-ui/pull/120/ will resolve this, because it introduces creating the Harvest object when the Harvest start job is run, and can be matched up later by id.

Issue #105, opened 2015-12-21
Harvest start message crashes twitter rest harvester


kerchner:
```
sfmtwitterrestharvester_1   | 2015-12-21 19:47:00,051: sfmutils.harvester --> Message is {
sfmtwitterrestharvester_1   |     "collection": {
sfmtwitterrestharvester_1   |         "path": "/tmp/collection/1", 
sfmtwitterrestharvester_1   |         "id": "1"
sfmtwitterrestharvester_1   |     }, 
sfmtwitterrestharvester_1   |     "id": 2, 
sfmtwitterrestharvester_1   |     "seeds": [
sfmtwitterrestharvester_1   |         {
sfmtwitterrestharvester_1   |             "token": "Kerchner", 
sfmtwitterrestharvester_1   |             "uid": ""
sfmtwitterrestharvester_1   |         }
sfmtwitterrestharvester_1   |     ], 
sfmtwitterrestharvester_1   |     "credentials": {
sfmtwitterrestharvester_1   |         "consumer_secret": "JvAUCmQNSLGSqKMbaKAR1NHyglAYovLTRHnWjAEZ4aQqhjSICB", 
sfmtwitterrestharvester_1   |         "access_token": "1934797142-dKCnhpuKnBEGJJ5jLUwArhEScHKr7ftJ9KHPUpR", 
sfmtwitterrestharvester_1   |         "consumer_key": "iHwtPZOQiJ0ML02rinhdhmEv2", 
sfmtwitterrestharvester_1   |         "access_token_secret": "yE44hYfBqVp2LiCEdOkh5pyOArAp9a9GjLnpQ9dNGIMp0"
sfmtwitterrestharvester_1   |     }, 
sfmtwitterrestharvester_1   |     "type": "twitter_search", 
sfmtwitterrestharvester_1   |     "options": {}
sfmtwitterrestharvester_1   | }
sfmtwitterrestharvester_1   | Traceback (most recent call last):
sfmtwitterrestharvester_1   |   File "twitter_harvester.py", line 85, in <module>
sfmtwitterrestharvester_1   |     TwitterHarvester.main(TwitterHarvester, QUEUE, [ROUTING_KEY])
sfmtwitterrestharvester_1   |   File "/usr/local/lib/python2.7/site-packages/sfmutils/harvester.py", line 419, in main
sfmtwitterrestharvester_1   |     harvester.run()
[...]

PR #106, opened 2015-12-23
refs #105. Makes id a str in message.

justinlittman:


kerchner:
Fixed in https://github.com/gwu-libraries/sfm-ui/pull/108 - closing PR

Issue #97, opened 2015-12-18
Update collection_detail along lines of wireframe


lwrubel:
* Note this includes a list of seedsets.
* Need to be able to edit information here.
* May need to create placeholder for changelog; changelog work should be ticketed separately?

kerchner:
For now, hard code SeedSet type choices next to the Add Seedset button.

PR #118, opened 2016-01-06
T95 collections list

### LINKED ISSUE #95: Update collection_list along lines of wireframe
Thinking that for the time being, we keep the "add collection" screen a separate view/template rather than, say, bootstrap modal as in wireframe.

lwrubel:
Basic collection list view with a few tests. Fixes #95

### DIFF
--- a/docker/app-dev/fixtures.json
+++ b/docker/app-dev/fixtures.json
@@ -1,70 +1,74 @@
 [
   {
-      "fields": {
-          "name": "testgroup",
-          "permissions": []
-      },
-      "model": "auth.group",
-      "pk": 1
+    "fields": {
+      "name": "testgroup",
+      "permissions": []
+    },
+    "model": "auth.group",
+    "pk": 1
   },
   {
-      "fields": {
-          "name": "testgroup2",
-          "permissions": []
-      },
-      "model": "auth.group",
-      "pk": 2 
+    "fields": {
+      "name": "testgroup2",
+      "permissions": []
+    },
+    "model": "auth.group",
+    "pk": 2
   },
   {
-      "fields": {
-          "username": "testuser",
-          "first_name": "",
-          "last_name": "",
-          "is_active": true,
-          "local_id": "",
-          "is_superuser": false,
-          "is_staff": false,
-          "last_login": null,
-          "groups": [1],
-          "user_permissions": [],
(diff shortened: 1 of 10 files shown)


kerchner:
- Cleanup:
  - _ui/urls.py_:  `reverse_lazy` is imported but unused.
  - _ui/views.py_:  PEP8 issues, and an import you can now do away with: `django.contrib.auth.models.User`
  - _ui/test_views.py_:  PEP8 issues, and an unused import: `django.contrib.auth.models.AbstractUser`
- Not sure whether this is within scope of #95 or not, but the UI for create collection allows me to select from all groups, not just groups in which I am a member.  (The Collections list, however, correctly shows only Collections owned by groups in which I am a member)
- The `success_url` for CollectionCreateView is `collection_list`.  Do you concur that CollectionUpdateView ought to redirect to the Collection list in the same fashion?
- _Authentication:_   Some of the views still need LoginRequiredMixin.  I verified that I'm able to access ui/collections/create/ - and successfully create a collection - as a user who is not logged in.  I can also update a collection at ui/collections/<number>/update/ .  
- Aesthetically and in the html, the bootstrap work looks right.  You may want to add padding between the Collections list and the Add New Collection button, and the page footer. 
- I see that you've created views for create/list/detail (but not update) Credentials, and I can access them, but there's no link to them from elsewhere in the UI.  I realize these views are beyond the scope of #95 - do you concur that we need additional ticket(s) for Credential views, and the appropriate place in the UI to link to them?

kerchner:
Looks like you fixed the PEP8 problems in the interim.

lwrubel:
Yes, sorry for not having taken care of that before making the pull request.

lwrubel:
I think most of the items you noticed have to do with scope of #95, which was just the ListView, as I understood it. 
* I'll work on the issue with selecting groups from the create collection page in ticket #96.
* Collection update success_url should probably be CollectionDetail, I'll work on that in ticket #97 
* I only put LoginRequiredMixin on CollectionListView since that was the scope of this ticket, should I go ahead and add it to all appropriate views at this time? I was just going to add it with the scope of each ticket, since I'd be writing tests for that at the same time. 
* Yes, we should ticket Credentials, I probably should not have worked on that in this ticket. I'm not sure what they should look like and what we need. I only created them to make it easy to have some data to work with in the collections flow. There may have been better ways to get that data into the database, via /admin or fixtures. I can remove this work if that's cleaner.

kerchner:
:+1:

PR #122, opened 2016-01-11
Add New Collection form and tests

lwrubel:


kerchner:
- Instead of setting `self.fields['name'].label = 'Collection name'`, use `verbose_name` in the model field.  We observed that crispy forms does pick up the verbose_name.
- In `CollectionForm`, `is_valid`, `full_clean`, `save` are not doing anything, so they should be removed. We can put back if we opt to override.

kerchner:
Refs Issue #96

kerchner:
@lwrubel on my instance, the Cancel button on /ui/collections/create/ doesn't do anything.  Does it work for you?

lwrubel:
The cancel button should be good now. Thanks for catching that.

PR #128, opened 2016-01-15
Collection detail view with seedset table. Fixes #97.

### LINKED ISSUE #97: Update collection_detail along lines of wireframe

lwrubel:
Note there are commented-out placeholders in the template for information which will be derived from the harvest status message.

### DIFF
--- a/sfm/ui/forms.py
+++ b/sfm/ui/forms.py
@@ -38,29 +38,36 @@ class CollectionForm(forms.ModelForm):
                 'group'
             ),
             FormActions(
-                Submit('submit', 'Add'),
+                Submit('submit', 'Save'),
                 Button('cancel', 'Cancel', onclick="window.history.back()")
             )
         )
+
+
+class AddSeedsetButtonForm(forms.Form):
+
+    collection = forms.ModelChoiceField(queryset=Collection.objects.all(), widget=forms.HiddenInput)
 
 
 class SeedSetForm(forms.ModelForm):
 
     start_date = forms.DateTimeField(required=False)
     end_date = forms.DateTimeField(required=False)
-
+ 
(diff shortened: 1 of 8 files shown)


kerchner:
- When creating a SeedSet, I would recommend passing the Collection ID via the request rather than in the URL.  Also, a URL like "ui/seedsets/<collection number>/create/" doesn't seem right; the number might be interpreted as the id for the new seedset.
- The Edit button is off in the upper right, which feels like a non-obvious place.  Is there a place it could reside that would be closer to where the focus of the page is, i.e. closer to the content?

lwrubel:
@kerchner the "add seedset" button form is passing the collection id via get. When I tried to do it via post, it tries to submit the seedset form. I'd welcome advice on this, I've probably been struggling too long on this ticket. Also, if I should squash and push again, let me know.

lwrubel:
@kerchner please re-review

kerchner:
- On collection edit page, the Add Seedset button is disabled because I assume we don't want the user leaving the collection details while they're open for editing.  However, the Seedsets in the Seedset list are clickable, and this would take the user off the collection edit page.  Do we want to consider not rendering the seedset names as links?
- Otherwise, looks good.

lwrubel:
I can go either way on that. I thought it might be useful for gathering
info that would inform the description or name edits. The links at least
don't go straight to a form, but to a DetailView. Maybe we could leave this
as is and then as we start using it, see how if it makes sense still?

kerchner:
:+1: Please squash, then I'll merge

Issue #136, opened 2016-02-04
tests for add seedset form


lwrubel:
Weren't included in #98, should be included in milestone 0.5.

kerchner:
@lwrubel determine whether this is already done - if so, close the ticket.

lwrubel:
Completed in 0.5.

PR #137, opened 2016-02-04
T98 seedset create page

### LINKED ISSUE #98: Update seedset_create along lines of wireframe
Harvest options depend on harvest type selected (which includes platform). 

For example, if starting with twitter harvester, provide options for
* twitter_search
* twitter_filter

lwrubel:
Fixes #98. Tests for form to be addressed in #136.

### DIFF
--- a/sfm/ui/admin.py
+++ b/sfm/ui/admin.py
@@ -59,18 +59,18 @@ class Collection(a.ModelAdmin):
 class SeedSet(a.ModelAdmin):
     fields = ('collection', 'credential', 'harvest_type', 'name',
               'description', 'is_active', 'schedule_minutes', 'harvest_options',
-              'max_count', 'stats', 'date_added', 'start_date', 'end_date')
+              'stats', 'date_added', 'start_date', 'end_date')
     list_display = ['collection', 'credential', 'harvest_type', 'name',
                     'description', 'is_active', 'harvest_options',
-                    'max_count', 'stats', 'date_added', 'start_date',
+                    'stats', 'date_added', 'start_date',
                     'end_date']
     list_filter = ['collection', 'credential', 'harvest_type', 'name',
                    'description', 'is_active', 'harvest_options',
-                   'max_count', 'stats', 'date_added', 'start_date',
+                   'stats', 'date_added', 'start_date',
                    'end_date']
     search_fields = ['collection', 'credential', 'harvest_type', 'name',
                      'description', 'is_active',
-                     'harvest_options', 'max_count', 'stats', 'date_added',
+                     'harvest_options', 'stats', 'date_added',
                      'start_date', 'end_date']
 
 
--- a/sfm/ui/forms.py
+++ b/sfm/ui/forms.py
@@ -1,5 +1,7 @@
 from django import forms
 from django.contrib.auth.models import Group
+from django.core.urlresolvers import reverse
+from django.utils import timezone
 from crispy_forms.helper import FormHelper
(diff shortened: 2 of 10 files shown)


lwrubel:
Note this includes migrations (max_count removed)

lwrubel:
@kerchner hold off on reviewing this. After consultation with @justinlittman, returning to the strategy of putting collection_id in the URL, seems supported better by django.

lwrubel:
@kerchner ready for review now

kerchner:
One test failure:
```
======================================================================
FAIL: test_seedset_harvest (ui.test_jobs.JobsTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/mock/mock.py", line 1305, in patched
    return func(*args, **keywargs)
  File "/opt/sfm-ui/sfm/ui/test_jobs.py", line 38, in test_seedset_harvest
    self.assertEqual("/test-data/collection/{}".format(self.collection.id), message["collection"]["path"])
AssertionError: '/test-data/collection/6' != '/sfm-data/collection/6'

----------------------------------------------------------------------
Ran 17 tests in 3.331s```

It looks like this test should use the configured `SFM_DATA_DIR` value rather than hard-coded `/test-data`

lwrubel:
@kerchner did you run with --settings=sfm.settings.test_settings ?

kerchner:
Apparently not.  Disregard earlier comment!

lwrubel:
no prob, I do this *all the time*

On Fri, Feb 19, 2016 at 1:31 PM, Dan Kerchner <notifications@github.com>
wrote:

> Apparently not. Disregard earlier comment!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/gwu-libraries/sfm-ui/pull/137#issuecomment-186352701>.
>



-- 


Laura Wrubel
Software Developer
George Washington University Libraries
2130 H St NW
Washington, DC 20052
202-994-6536
lwrubel@gwu.edu

kerchner:
- In models.py:  `SeedSet.start_date` has a helpful `help_text`. Was it a conscious choice not to include similar `help_text` for `SeedSet.end_date`?
- Field validations of the dates work well.   I see that start_date cannot equal end_date.  What if I just want to run the collection for a few hours tomorrow?
- Just noting that SeedSet names are not required to be unique - only ids are.
- Harvest type should be a required field.  It is not currently.
- Consider adding a test similar to`test_collections_detail_anonymous` but for the seedset create URL.  Or we could have single test method that tests each of the views which should require login.

lwrubel:
* I can add an end_date help_text, wasn't initially thinking it was needed, but someone did ask about it in our project meeting. Is my assumption correct that if it is blank, the harvests continue indefinitely?
* Will add a test for login, noting that we should have additional tests for the forms. Will work with Justin, within the next milestone scope, to understand and implement further tests for this area.

lwrubel:
Note that start_date and end_date supporting entering times (24-hour clock), so it is possible to set up a seedset with a start and end date on the same day. We should ticket separately adding a calendar / time widget to this field to make that clearer and easier.  Bootstrap3 doesn't already have that, but there are other options to plug in. 

The seedset model currently has the harvest_type field with blank=True, allowing a blank field to validate. I'll removed that to make this field required. If there was a reason we initially allowed blank values, let me know.

kerchner:
@lwrubel :+1: although you may want to clean up PEP8 compliance in `sfm/ui/test_views.py`.  Please go ahead and squash, then we can merge the branch.

PR #147, opened 2016-02-25
Modified harvest jobs to ignore inactive seeds and seedsets. closes #138

### LINKED ISSUE #138: Is active flag ignored when scheduling harvests
When scheduling a harvest (sched.schedule_harvest), the harvest should only be scheduled if the seedset is active.  Otherwise, it should be unscheduled.

When executing a harvest (jobs.seedset_harvest), a harvest request message should only be sent if there are active seeds.

rajatvij:
Please check if it is working properly and let me know if i need to change anything.


justinlittman:
Please add tests to test_sched.py and test_jobs.py.

Issue #121, opened 2016-01-07
set up static files


lwrubel:
specific first need is in order to add css to improve vertical button spacing as in #95 and #96

rajatvij:
@kerchner and @justinlittman According to two scoops book. Both static files and templates needs to be at same folder where manage.py is so all apps can have a single place for all templates and static files instead of each app having their individual templates and static files. Should i follow that or i can place static folder inside ui folder which is more in sync with what we are currently having. Which option sounds better?

Option 1:
- sfm-ui/
  - sfm/
     - message_consumer/
     - sfm/
     - ui/
     - templates/
     - static/
     - manage.py

Option 2(i prefer this):
- sfm-ui/
  - sfm/
     - message_consumer/
     - sfm/
     - ui/
         - templates/
         - static/
     - manage.py

Option 3:
Or i can leave templates folder where it is but i can create static folder at the same level as manage.py

kerchner:
sfm-ui is not likely to entail more than one app, so Option 2 should be fine.

PR #149, opened 2016-02-29
Created a static folder to support static files in sfm-ui. refs #121

rajatvij:
Please check and let me know if any modification is required for the branch.
The static files are just to create folders and should be deleted later.

### DIFF
--- a/sfm/ui/static/ui/css/test.css
+++ b/sfm/ui/static/ui/css/test.css
@@ -1,0 +1,22 @@
+body {
+    font-family: "Trebuchet MS", Verdana, sans-serif;
+    font-size: 16px;
+    background-color: dimgrey;
+    color: #696969;
+    padding: 3px;
+}
+
+#main {
+    padding: 5px;
+    padding-left:  15px;
+    padding-right: 15px;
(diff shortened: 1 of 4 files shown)


sfm/sfm/settings/common.py
```diff
@@ -125,7 +125,7 @@
 # Static files (CSS, JavaScript, Images)
 # https://docs.djangoproject.com/en/1.8/howto/static-files/
 
-STATIC_URL = '/static/'
+STATIC_URL = '/ui/static/'
```
justinlittman: This puts all of the static files in /ui/static/, including the admin files.  I think you  want to leave this as /static/ and then put the sfm-ui static files in sfm/ui/static/ui/ instead of sfm/ui/static/.

See https://docs.djangoproject.com/en/1.9/howto/static-files/

rajatvij:
@justinlittman You were right there was no need for '/ui/static/'. Even before it was working with '/static'. I have made changes as you asked. Just to make it clear. Now to include static files we would need to refer static files in template like : {% static 'ui/css/test.css' %}

Issue #146, opened 2016-02-25
Credential UI design mockups


kerchner:
Supports #133.

- List view for list of my credentials, w/ability to Add a new credential
- Create credential screen
- Update credential

rajatvij:
@lwrubel @kerchner Please check following wireframe for the Credentials UI Screen, Let me know if any modification is required in screen of if we dont need changelog there.

https://drive.google.com/drive/u/0/folders/0B6mKHHXuA0dfZlJmcUI2SkVUOEU

kerchner:
Closing; #133 is occurring already, no need to go back and create wireframes.

PR #150, opened 2016-02-29
T133 sfm credentials ui, closes #133

### LINKED ISSUE #133: Credential UI
- Designs for UI
- Create tickets for UI work

rajatvij:
Updated Credentials UI screens to save twitter and flickr credentials. To test, we would need to goto credential list screen which is /ui/credentials/ in url.

I haven't created a link for credentials yet as i didn't know where to put it for now.

Please check and let me know if this is right approach, if not then i will update the #146 and try to get this as close as possible to wireframe.


rajatvij:
@lwrubel I have added screen to add weibo's credentials as discussed in today's meeting along with a link to view credential list on navbar to navigate easily. Please check and let me know if you find any issue with it.

lwrubel:
Since we had no wireframes for this in advance, some of the UI needs only became clear now that we have screens to review. A few requested changes while I'm still reviewing:
- We need some way to identify each credential that's more friendly than the __str__ output. Please add a name field to the Credential model and run makemigrations to create a migrations file. 
- Follow the UI patterns established in the Collections and Seedset pages. I have some marked up printouts I'll send you. Specifically:
  - Credentials (credentials_list) should be similar to My Collections, with a table first and then options to create new ones. No heading needed on the list. Move the options to add new credentials below the table.
  - Add breadcrumbs to all pages as on Collections and Seedset pages. Unlink page headings (breadcrumb replaces that navigation)
  - Credential Detail should look like Seedset Detail with label/value instead of a bulleted list. Update page to have the new name field instead of __str__
  - Move Change Note to bottom of forms. 
  - Use the same styling for the Create and Cancel buttons, and use "add" instead of "create".
- Platform field should not be editable in Update view. Just show it as text on the screen.
- It looks like is_valid and full_clean in the forms are not doing anything beyond the default django form handling, so I think you can remove those for now.

lwrubel:
Some UI edits:
* With breadcrumbs, you don't need to repeat the last item. The page you're on should be the last item listed, and it doesn't need to be a link. 
* Typo in credential_detail.html: says Is Active:: (remove extra colon)
* When you're on the credential_create page for each platform, there isn't anything to tell you what platform you're working with. Could you add that to the heading at the top of the page?
* The Add New Weibo Credential form has "history note" whereas the others have "Change note". 

As discussed, there is a delete view/form, as I understood it we'll add delete buttons and the supporting UI/function later (e.g. what happens to a seedset when you delete its credential?). That should go into a new ticket (@kerchner to confirm).

kerchner:
Closing without merge; superseded by #180 .

Issue #54, opened 2015-11-09
Keep track of changes to collections, seed sets, seeds


justinlittman:
https://github.com/etianen/django-reversion?

justinlittman:
https://www.djangopackages.com/grids/g/model-audit/

justinlittman:
I'd take a closer look at these:
* https://github.com/treyhunner/django-simple-history
* https://github.com/jjkester/django-auditlog

justinlittman:
So like django-reversion, django-simple-history records historical versions of model objects.  This is optimized for reverting to a previous version of a model object.  However, it is not optimized for our use case of wanting to provide a log of changes.  (We would be responsible for comparing versions and determining what has changed.)

dchud:
It seems possible that both goals might be worthy features - an "undo" based on -reversion or -simple-history might be something users are glad to have.  But, yes, the goal of this ticket is to log changes for the sake of record keeping.

justinlittman:
A few notes:
* Until https://github.com/jjkester/django-auditlog/pull/47 is accepted, referring to our fork of django-auditlog.
* Formatting of the history is performed in `templates/ui/history_snippet.html`. This approach can be changed depending on what makes the most sense for the UI.
* A note field is added to the collection, seed, and seedset model.  This is a sneaky way of adding a note to the log entry.  The note field in model instances should be ignored.  (In the model update pages it is set to blank.)  When formatting the log entry, the note can be extracted and handled appropriately.

justinlittman:
Per discussion with @kerchner and @lwrubel, leaving not merging these commits until later this milestone or early next.

justinlittman:
For #117, this will need to be added to sfm_ui_consumer.py:

When seeds are updated based on tokens that have changed:
```
                       seed.note = "Updated token based on harvest {}".format(self.message["id"])
```

When seeds are updated based on uids that have been found:
```
                        seed.note = "Updated uid based on harvest {}".format(self.message["id"])
```

justinlittman:
Also, in test_sfm_ui_consumer.py when checking updated seeds:
```
        self.assertTrue(seed1.note)
```
and
```
        self.assertTrue(seed2.note)
```

kerchner:
Resolved in https://github.com/gwu-libraries/sfm-ui/pull/151

PR #151, opened 2016-03-02
refs #54. Adds history support to Credentials, Collections, SeedSets,…

justinlittman:
… and Seeds. This includes tracking history and displaying on detail page. Also, hooks harvest into history to record the Credentials, SeedSets, and Seeds used to request the harvest.


justinlittman:
To test:
1. Create and change credentials, collections, seed sets, and seeds in the UI or Admin. Note the addition of a note field.
2. Go to the details page for credentials, collections, seed sets, and seeds. Note the addition of a list of changes to the object.  This list should reflect the changes made to each field of the object, the time of the change, the user that made the change, and the note (if provided).
3. Go to Admin. Note the addition of historical credentials, collections, seed sets, and seeds.

kerchner:
Making no changes to an object, but saving, results in a new Historical object.  Is the correct behavior when there are no diffs to any fields in `diff_fields` ?

justinlittman:
It is.  `diff_fields` is something I added and is only used to determine which fields to display diffs for.

kerchner:
What would it take for us to avoid creating a Historical object if this was a no-change save?  @lwrubel suggests perhaps using validation (on the ui side) to not save if we determine that the object hasn't changed.

justinlittman:
@kerchner Made the suggested change so that only creates historical record when diff field is changed.

kerchner:
@justinlittman The code certainly looks right, but no-change saves are still creating new Historical objects.

Also while you're in there, would you consider a slightly more descriptive value for `HISTORY_NOTE_LABEL` than `"Note"`?  Granted, there is `HISTORY_NOTE_HELP`, but perhaps something like `"Change Note"` would be even clearer.

kerchner:
@justinlittman Disregard.   `history_save` is working correctly.

kerchner:
Please proceed with squash (and update `HISTORY_NOTE_LABEL` if you agree)

justinlittman:
@kerchner - Good to go.

PR #153, opened 2016-03-11
T148 remove collection is active

lwrubel:


justinlittman:
I'm working on #138, so grabbing this PR as well.

Issue #164, opened 2016-04-03
Git apt package version needs update


kerchner:
This, and errors like it, will occur less frequently with the resolution of #141.

`Version '1:1.9.1-1ubuntu0.2' for 'git' was not found`

Available `git` version is now `1:1.9.1-1ubuntu0.3`

This is preventing successful setup of new docker container for sfm-ui.  Will resolve this by fixing and merging #141.

kerchner:
Resolved with PR #165

PR #165, opened 2016-04-03
T141 Implements limited apt package wildcarding

kerchner:
To test:

Temporarily substitute the URL
`https://raw.githubusercontent.com/gwu-libraries/sfm-ui/t141-package-wildcarding/requirements/requirements.apt` within the `ADD` directive near the top of your Dockerfile which gets `requirements.apt` and places it within the container.

Build in a completely clean environment (meaning, remove all docker images first).

If successful, you should not see errors like the error described in #164.


kerchner:
Also fixes #164.

PR #180, opened 2016-04-07
Created and Updated Credentials Screens according to the wireframes and current UI. closes #133

### LINKED ISSUE #133: Credential UI

rajatvij:
@lwrubel  New branch for Credential UI screens for clarity of changes and to deal with merge conflicts. 
Please check this branch for Credentials UI and ignore the other one.

### DIFF
--- a/sfm/ui/forms.py
+++ b/sfm/ui/forms.py
@@ -5,7 +5,6 @@ from django.utils import timezone
 from crispy_forms.helper import FormHelper
 from crispy_forms.layout import Layout, Fieldset, Button, Submit
 from crispy_forms.bootstrap import FormActions
-
 from .models import Collection, SeedSet, Seed, Credential
 
 HISTORY_NOTE_LABEL = "Change Note"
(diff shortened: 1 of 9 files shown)


lwrubel:
Apologies for not catching this before, but I'm noticing that if I log in as testuser2, I see the credentials entered by testuser. CredentialListView should have logic to show the right credentials based on the user. 

We will need to figure out the situation where a user who is part of a group sees collections in their group, but which they did not create: what aspects of the credential should the be able to see, and to edit? The name, or any field?

rajatvij:
We might need to discuss this a bit more as i am not sure if a user should be able to edit any credentials in the group or only able to view them but edit only those that he created. It is more of an issue of permission than what should be displayed. As of now, i think django inherently provide access to the objects of group that you are part of if you have set the permissions of your user that way, hence we can see and even edit the credentials of the group.

@kerchner Please correct me if i am wrong in my assumption above. And do we need a new ticket regarding user group and permissions?

kerchner:
I created #184 for the scope of working through this question, deciding, and implementing filters on who can view/edit/use other users' credentials - if at all, and when.

@rajatvij and @lwrubel for now let's merge this without filtering, as it achieves the basic purpose of putting a UI on credential creation.  Please let me know if you have strong objections.

lwrubel:
Since this is documented, and assuming it's to be done by 1.0, I don't object.

Issue #145, opened 2016-02-25
Change harvest messages to use seed id


justinlittman:
Seeds now have ids.  Those ids should be passed around in the harvest messages to reduce ambiguity.

justinlittman:
More details:
* In an harvest start message, each seed should have an `id` key containing the seed id.
* In a harvest status message, `token_updates` and `uids` should use ids as keys.

Thus, this will require changing the base harvester (in sfm-utils), as well as `jobs.py` and `sfm_ui_consumer.py` (in sfm-ui).

PR #204, opened 2016-04-20
Adds seed_id in harvest messages. refs #145

rajatvij:


sfm/ui/jobs.py
```diff
@@ -60,21 +63,29 @@ def seedset_harvest(seedset_id):
...
             seed_map = {}
+            seed_map["seed_id"] = historical_seed.seed_id
             if historical_seed.token:
                 seed_map["token"] = historical_seed.token
             if historical_seed.uid:
                 seed_map["uid"] = historical_seed.uid
             message["seeds"].append(seed_map)
 
-    routing_key = "harvest.start.{}.{}".format(historical_credential.platform, harvest_type)
+    # Include conditions to check Is Active Flag here.
```
justinlittman: Is adding this necessary given that there is an is active check at the beginning of this function?

justinlittman:
In the message, the key should be "id" not "seed_id" to be consistent with other parts of messaging spec.  Also, the messaging spec needs to be updated as well.

rajatvij:
@justinlittman You were right about the redundant seed_set active condition. I will change seed_id to id in messages shortly. For messaging spec, Do i just need to edit https://github.com/gwu-libraries/sfm-ui/blob/master/docs/messaging_spec.rst?

justinlittman:
Yes.

rajatvij:
@justinlittman Please check this with https://github.com/gwu-libraries/sfm-twitter-harvester/pull/8 and https://github.com/gwu-libraries/sfm-flickr-harvester/pull/6. This is up for testing. Apologies for multiple commits. Let me know if there are any issues.

Issue #181, opened 2016-04-08
Support seedless harvest types


justinlittman:
Some harvest types (e.g., Twitter stream and Weibo friends) don't have seeds.  Since a seed is required to send a harvest start message, the current kluge is add a dummy seed, which is then ignored by the harvester.

To fix this:
* Create a means of identifying seedless harvest types.
* When a harvest submit is triggered, if the harvest type is seedless then skip the check that requires a seed.
* Send the seed set identifier in the harvest start message.
* Change the UI so that seeds cannot be added for seed sets of seedless harvest types.
* Change the UI so that seeds or seed counts are not listed for seed sets that are seedless harvest types.
* Update the message spec to reflect the above.

kerchner:
Fixed in #222

Issue #159, opened 2016-03-28
Activate/inactivate for seeds, seed lists


kerchner:
The seeds list on the seedset page should allow the user to inactivate/activate individual seeds, and add a change note to these changes.

We should discuss and make a decision on whether to provide for deletion of seeds.

justinlittman:
Deleting seeds complicates history, so I would encourage inactivation.

kerchner:
Spun off #185 to deal with the question of deletion.

kerchner:
- A Seed will have an active/inactive toggle
- A SeedSet will have a "Pause"/"Recording" (or similar) button (which uses the existing is_active flag) and does not modify any SeedSet field values (e.g. frequency, etc.)

justinlittman:
Here's a proposal that may simplify and allow consistency across streaming and non-streaming harvests.
1.  Get rid of harvest start date.  (Can you think of a significant use case for this?)
2.  When the stop date for a harvest passes, it is toggled to pause.
3.  Changes to a seedset or a seed or the addition of a seed to a seedset can only be made when the seedset is paused.
4.  Seedsets are paused by default when they are created.

So, for a non-streaming harvest type:
1. User creates the seed set.  (Seed set is paused and editable.)
2. User adds seeds to the seed set.
3. User unpauses the seed set.
4. Harvests are scheduled.
At a future point:
1. User pauses the seed set.  (Seed set is editable.)
2. User adds a new seed and inactivates an existing seed.
3. User unpauses the seed set.
4. Harvests are unscheduled.
When the scheduled stop time is reached:
1. Harvests are unscheduled.
2. Seed set is paused.

For a streaming harvest type:
1. User creates the seed set.  (Seed set is paused and editable.)
2. User adds seeds to the seed set.
3. User unpauses the seed set.
4. Harvest start message is sent.
At a future point:
1. User pauses the seed set.  (Seed set is editable.)
2. Harvest stop message is sent.
2. User adds a new seed and inactivates an existing seed.
3. User unpauses the seed set.
4. Harvest stop message is sent.
When the scheduled stop time is reached:
1. Harvests stop message is sent.
2. Seed set is paused.

Let's discuss.

justinlittman:
@lwrubel Are you OK with me taking this ticket and proceeding?  I'd also like to knock off #181 and #129 at the same time.

lwrubel:
[...]

Issue #129, opened 2016-01-15
Handle streaming jobs


justinlittman:
As noted in the [messaging spec](http://sfm.readthedocs.org/en/latest/messaging_spec.html#flow), streaming harvests have a different flow than REST harvests.  Streaming harvests require that start and stop messages be sent.

This ticket is for adding support for this to SFM UI.

kerchner:
Ref. #129.   Pause/Record button on SeedSet could be the UI control that triggers start/stop messages to a streaming harvest.

PR #222, opened 2016-04-28
T159 toggle active

justinlittman:
Fixes for #159, #181, #129.


justinlittman:
Here's some items to consider testing:
* Make sure that start data has disappeared.
* Create a seed set that can have multiple seeds (e.g., twitter user timeline). Verify that seed set cannot be started until at least one seed is added.  Verify that seed set or seeds can only be edited when harvesting is paused.
* Create a seed set that can only have a single seed (e.g., twitter search). Verify that seed set cannot be started until there is a seed.  Verify that more than one seed cannot be added.
* Create a seed set that can have no seeds (e.g., weibo). Verify that a seed cannot be added and that seed set can only be edited when harvesting is paused.
* Create a stream seedset (e.g., twitter sample). Verify that a harvest start message is sent when harvesting is started (well, after waiting 15 seconds). Verify that a harvest stop message is sent when harvesting is stopped.  Verify that a harvest stop message is sent when harvest end date has passed.

kerchner:
@justinlittman upon turning my attention back to this PR, I see that it now needs rebasing.

kerchner:
@justinlittman I verified the items above, and also like how the SeedSet is not editable while harvesting is active.

However, I was able to add 2 seeds to a Twitter Search seedset, which should only accommodate a single seed (`Verify that more than one seed cannot be added`)

kerchner:
It looks like perhaps you meant to add `TWITTER_SEARCH` to `SeedSet.REQUIRED_SEED_COUNTS` ?

justinlittman:
I did and caught that later on.  It is fixed in
https://github.com/gwu-libraries/sfm-ui/commit/c353f987a2a59f5c840a748decd4986e7f083d99,
which is in a different PR.

justinlittman:
@kerchner I'm making the presumption that it is then OK to merge this PR.  I'm going to proceed so that I can get the next PR rebased and ready to go.  Hoping this is acceptable.

kerchner:
:+1:

Issue #207, opened 2016-04-25
Notifications for export completion


lwrubel:
Add in-ui support for notifications when export complete. Consider email also.

rajatvij:
We can use following approach for this:

1. Use default [django messages](https://docs.djangoproject.com/en/1.8/ref/contrib/messages/) for flash user notifications and Use default [django send_mail()](https://docs.djangoproject.com/en/1.8/topics/email/) for sending mails. 
NOTE: Both django messages and send_mail() are straightforward and can be implemented without much work. This approach might be simpler to implement.

2. Use third-party apps such as [django-heythere](https://django-heythere.readthedocs.org/en/latest/index.html#indices-and-tables) for doing both together. 
NOTE: There are many apps for notification which supports from tweets to sms. The given app supports both mail and django messages and pretty straight forward. The only advantage here would be that we would need to change code similar to django messages and it is persistent.

Let me know which sounds better.

justinlittman:
Please proceed with the first approach.

rajatvij:
@justinlittman I am unable to send mail via google as they block sending mails from lesser secure apps and sends a sign-in prevented email response.

Error Message

> 
smtplib.SMTPAuthenticationError: (534, '5.7.14 <https://accounts.google.com/signin/continue?sarp=1&scc=1&plt=AKgnsbsh\n5.7.14 eRpxfaUIj_83NvtOhXY-5grgnKfkNR6w_71axh0sfA3rmEGgTTQzlVgLcsDkzhwGti3-MS\n5.7.14 Ra_JMi_xfR6Gy3ATwdNqQIjAbPyOyBwtKQU6SJBjWFSRSGSd3jCD42gXHvsIeCtPLncHaX\n5.7.14 INDplsS_ywcYNtdDaxoPG7HfoJgqBTpPj2QLohtKSIJ2g-PB6nVuosp2rP1tMB3ZS9UqBb\n5.7.14 nftsuwy-CzgsHK2E-ExZNYeKQ-Pbk> Please log in via your web browser and\n5.7.14 then try again.\n5.7.14  Learn more at\n5.7.14  https://support.google.com/mail/answer/78754 f91sm3319672qgf.4 - gsmtp')

The way we can bypass this is by allowing lesser secure app to use the email id from which we need to send the mail to users by Turning on following security feature https://www.google.com/settings/security/lesssecureapps. 

[...]

PR #225, opened 2016-04-29
Adds mail notification on export. refs #207

rajatvij:
Steps to Test:

1. Edit settings/common.py with your Email ID and Password along with its smtp server
2. Log in to sfm-ui and edit user detail in admin with the email id on which you wish to receive email notification.
3. Perform an Export.

NOTE: I have observed a delay of 3 to 6 minutes before we get email notification.

### DIFF
--- a/sfm/message_consumer/sfm_ui_consumer.py
+++ b/sfm/message_consumer/sfm_ui_consumer.py
@@ -2,9 +2,14 @@ import logging
 from sfmutils.consumer import BaseConsumer
 from ui.models import Harvest, SeedSet, Seed, Warc, Export
 import json
+from django.core.mail import send_mail
+from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
+from django.core.urlresolvers import reverse
+from django.contrib.sites.models import Site
 import iso8601
 import time
+from smtplib import SMTPException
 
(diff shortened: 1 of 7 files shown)


sfm/message_consumer/sfm_ui_consumer.py
```diff
@@ -102,13 +104,24 @@ def _on_export_status_message(self):
             log.debug("Updating export with id %s", self.message["id"])
             # Retrieve export model object
             export = Export.objects.get(export_id=self.message["id"])
-
+            # Retrieve user model object
+            user = User.objects.get(username=getattr(export,'user'))
+            # Get reciever's email address
+            reciever_email = getattr(user, 'email')
```
justinlittman: Is there a reason you're using getattr instead of just calling the attribute, e.g., `export.user`?

Receiver is spelled "ei" instead of "ie".

justinlittman:
@lwrubel Can you author the email that should be sent to users on exports?

@rajatvij The email will need to include a link, so you may want to start figuring out how to generate it.

rajatvij:
@justinlittman Yes you were right, I have updated the code as per your comments and have commented a way to format url in the mail message as i was not sure on what url we would be forming or sending in the mail. Please let me know if you need me to change or add anything to this.

Issue #224, opened 2016-04-29
Simple stats rollup


justinlittman:
Based on stats in harvest status messages, collect a display stats for seed set and collection.

kerchner:
Resolved in https://github.com/gwu-libraries/sfm-ui/pull/228

PR #228, opened 2016-05-02
refs #224. Adds stats rollup and display for seed sets and collections.

justinlittman:


justinlittman:
To test:
1. Create and start harvests of one or more seed sets.
2. Wait for harvests to be performed.
3. Observe counts of items harvested on collection and seed set detail pages.

justinlittman:
@rajatvij - Sorry, but I realized that the stats fields were never changed to JSONFields.

rajatvij:
@justinlittman The stats that i am able to see for individual harvest types are as follows:

Flickr User harvest - users and photos
Twitter timeline - tweets and web resources. Are not expecting photos here?
Twitter Search - Nothing. Are we not expecting anything for twitter search?

Please let me know if i should expect something else.
As far as the code related to the ticket is concerned, this seems to be working fine as it is displaying all stats from summary section of harvest return message. I just want to double check on what should be returned in summary section of each harvest success message.

justinlittman:
Good point.  The names of the stats items is confusing.  I'm going to clarify (e.g., "Flickr photos" instead of just "photos".)

You should see tweets for Twitter search.  I'll test and see if I can replicate -- I suspect you may have just had a problem with your search.

justinlittman:
https://github.com/gwu-libraries/sfm-flickr-harvester/commit/9f2993f8ea4f48e74a20935c4d5c079f21cd634b tweaks the stats for Flickr.

justinlittman:
I've verified that search correctly returns tweets.  I'm going to go ahead and rebase/merge.

PR #231, opened 2016-05-02
T220 hide request export button. closes #220

### LINKED ISSUE #220: Request export button not display until there is a successful harvest
This avoids a situation where an export fails because there are no harvests completed and/or no WARCs yet.

rajatvij:
This disables the "Request Export" button on Seedset detail screen where harvest for the seedset is incomplete, failed or has not started yet.

Steps to Test:
1. Log in to sfm-ui, create credentials, and collection. Add seedset to the collect.
2. Go to seedset details screen and check if the request export button is disabled.
3. Add seeds to the seedset and update seedset to initiate a harvest or initiate a harvest from admin.
4. After a minute when the harvest is completed. Check if the harvest's status is completed successful. If so, then the "Request Export" button should be enabled on Seedset details screen.

### DIFF
--- a/docs/messaging_spec.rst
+++ b/docs/messaging_spec.rst
@@ -280,7 +280,8 @@ created during a harvest. Example::
             "id": "test_collection"
         }
         "harvest": {
-            "id": "sfmui:45"
+            "id": "sfmui:45",
+            "type": "flickr_user"
         }
     }
 
--- a/sfm/ui/admin.py
+++ b/sfm/ui/admin.py
@@ -39,10 +39,10 @@ class UserAdmin(AuthUserAdmin):
 
 
 class Credential(a.ModelAdmin):
-    fields = ('user', 'platform', 'token', 'is_active', 'date_added', 'history_note')
-    list_display = ['user', 'platform', 'token', 'is_active', 'date_added']
-    list_filter = ['user', 'platform', 'token', 'is_active', 'date_added']
-    search_fields = ['user', 'platform', 'token', 'is_active', 'date_added']
+    fields = ('user', 'platform', 'name', 'token', 'is_active', 'date_added', 'history_note')
+    list_display = ['user', 'platform', 'name', 'is_active', 'date_added']
+    list_filter = ['user', 'platform', 'is_active', 'date_added']
+    search_fields = ['user', 'platform', 'name', 'is_active', 'date_added']
 
 
 class HistoricalCredential(a.ModelAdmin):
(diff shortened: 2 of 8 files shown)


lwrubel:
@rajatvij When I click on a seedset from a collection_detail view, I get an error that seems to go back to this line: 
`if Harvest.objects.get(harvest_type=self.object.harvest_type).status == "completed success":
`

The error is something like: 
"get() returned more than one Harvest -- it returned 902!"

Would filter() accomplish what you need here? I think what you're really trying to do is see if at least one has status == "completed success" 

While you're in there working on this, you can simplify the template if/else logic by just putting the following in the button HTML:
`{% if not can_export %} disabled {% endif %}
`

rajatvij:
@lwrubel I have fixed the issue, please check and let me know if any other changes are required.

lwrubel:
@rajatvij, the user should be able to request an export as long as there is one "completed success" harvest. Right now, the status of the last harvest is what is determining the availability of the request export button. I have a seedset with many harvests, some of which have completed successfully, but because the more recent ones have not completed, I don't get the button.

How about instead, checking to see if there is at least one "completed success"?  You could break out of the loop once one is found, so you don't have to check every single harvest.

rajatvij:
@lwrubel sorry, i misinterpreted the ticket before. I have corrected the condition now. Please check and let me know if you find any other issue with the ticket.

kerchner:
@lwrubel It looks like the merge for #236 may have caused merge conflicts that need to be resolved.

lwrubel:
@rajatvij, would you review the merge conflicts and fix?

rajatvij:
@lwrubel I have rebased the branch with master to fix merge conflicts.

lwrubel:
@rajatvij It looks like that old line of SeedSetUpdateView snuck into views.py. Would you replace 
`context["collection"] = Collection.objects.get(pk=self.kwargs["collection_pk"])`
with
`context["collection"] = Collection.objects.get(pk=self.object.collection.pk)`
then squash?

lwrubel:
@rajatvij when I try to edit a seedset, I still get a KeyError because of line 194 of views.py 
`kwargs['credential_list'] = _get_credential_list(self.kwargs["collection_pk"], self.kwargs["harvest_type"])`

rajatvij:
@lwrubel Sorry, I introduced that issue while resolving merge conflicts. I have corrected that, this should be ok to merge now.

PR #236, opened 2016-05-05
T212 harvest options

justinlittman:


justinlittman:
This is done but will need to be rebased once earlier are merged.

justinlittman:
@kerchner Rebased and ready to go.

PR #267, opened 2016-05-18
Harvest stats by time and home page visualizations

justinlittman:


justinlittman:
@lwrubel Hold off on testing this.  Just realized that I'm not handling harvests that span days well.

justinlittman:
@lwrubel. This is ready to go.  You'll need to pull master for sfm-utils and all of the harvesters.

lwrubel:
@justinlittman I think it's ready to go. Want to squash these commits before I merge?

Issue #279, opened 2016-05-23
Tweaks on harvester list pagination 


Tanych:
When the number of harvester list page becomes large, it's very hard to locate the last page. It's necessary to add the first page and last page direction in the harvester pagination.

Looks like:
Previous 1 of 25 Next  »
« Previous 4 of 25 Next  »
« Previous 25 of 25 Next

Also, if possible, we can separate the pagination page in order to make the changes to all the similar issues.

PR #280, opened 2016-05-23
Tweaks on harvester list pagination

Tanych:
The details of the issues #279


justinlittman:
@Tanych Please also make this change for the change log pagination as well.

Issue #208, opened 2016-04-25
Rename "seedset"


lwrubel:
Current term "seedset" does not work for harvests that don't have seeds. Replace with something more universally applicable.

kerchner:
@lwrubel Add candidate terms (so far) from Slack conversation.

kerchner:
This needs to be changed on a very stable code base.  Isolating as its own milestone.

lwrubel:
Possibilities:
harvest set, subcollection, collection part, collector, harvest strategy, harvest definition, harvest plan, job, job plan, pipe, net, map, funnel

lwrubel:
Will go forward with the terminology:

Collection Set > Collection > [optional: Seed]

justinlittman:
PRs for this ticket:
* https://github.com/gwu-libraries/sfm-ui/pull/293
* https://github.com/gwu-libraries/sfm-utils/pull/14
* https://github.com/gwu-libraries/sfm-flickr-harvester/pull/7
* https://github.com/gwu-libraries/sfm-twitter-harvester/pull/12
* https://github.com/gwu-libraries/sfm-weibo-harvester/pull/15

lwrubel:
Closed with #293 .

PR #293, opened 2016-06-01
refs #208. Renames collection to collection set and seedset to collec…

justinlittman:
…tion.


lwrubel:
ChangeLogView in `views.py` uses apps.get_model and it's expecting the exact model name. 
```
ModelName = apps.get_model(app_label="ui", model_name=model_name)
```
So /ui/collection_set/2/changes/ does not work but /ui/collectionset/2/changes/ does work. I don't think we use this anywhere else, so we could change the view to handle collection_set, rather than change urls.

lwrubel:
Or change the model_name passed in the context data.

lwrubel:
I'm seeing some seedset references in the docs, specifically messaging_spec.rst and processing.rst, and harvester.rst might need an quick check of the WARC path.

justinlittman:
@lwrubel Can you double check that you're looking at the right branch for the docs?  I'm not seeing any errant seedset references.

lwrubel:
@justinlittman You're right, I was looking at the wrong branch. Sorry about that.

lwrubel:
@justinlittman that latest commit fixes getting to the change log page for a collection set, but the breadcrumbs in the change_log.html template are looking for "collection_set" and are not displaying. 

The h2 should probably also say "collection set" rather than "collectionset". But that can go in another ticket if needed.

Issue #291, opened 2016-05-28
Restarting sfm-ui container running apache fails


justinlittman:
Steps to reproduce:
1. Bring up an instance of an sfm-ui container that uses apache (i.e., not dev-runserver).
2. Kill the container the `docker kill`.
3. Bring the container back up with `docker-compose up -d`.

Expected result:
Apache starts back up.

Actual result:
Apache does not start back up.

Work around:
Remove the container with `docker rm -v`.

Tanych:
## Information
* Code related to this issue: [invoke.sh](https://github.com/gwu-libraries/sfm-ui/blob/master/docker/app-dev/invoke.sh)
``` bash
echo "Running server"
export SFM_RUN_SCHEDULER=True
source /etc/apache2/envvars
apachectl -DFOREGROUND
```
* Differences among `docker kill`, `stop` and `rm -v`. Ref: [here](https://www.ctl.io/developers/blog/post/gracefully-stopping-docker-containers/)
* Running apache on foreground: http://zroger.com/blog/apache-in-the-foreground/

## Test Cases
### 1. Using `docker kill` or `docker stop` 
* Number of test samples: 10
* Results: `docker_sfmdevapp_1` exited with 0
* Error info: The httpd pid number could be different.
```python
sfmdevapp_1               | AH00558: apache2: Could not reliably determine the server's fully qualified domain name, using 172.17.0.5. Set the 'ServerName' directive globally to suppress this message
sfmdevapp_1               | httpd (pid 77) already running
docker_sfmdevapp_1 exited with code 0
```
* Reason: Sometime, the `docker kill` or `docker stop` can't stop the apache on the contanier, when the sfmdevapp docker restart with running `apachectl -DFOREGROUND`, it will fail since the httpd sevice was already running.

### 2. Change the apache running 
modify the `invoke.sh` like this:
```python
echo "Running server"
export SFM_RUN_SCHEDULER=True
source /etc/apache2/envvars
/etc/init.d/apache2 start
sleep 1
/etc/init.d/apache2 graceful-stop
sleep 1
echo "start apache on foreground"
apachectl -DFOREGROUND
```
* Number of test samples: 10
* Results: `docker_sfmdevapp_1` restart successfully
* Analysis: If only adding `/etc/init.d/apache2 graceful-stop`, sometime it works. 
* Problem: whether it could re-cause this issue: https://github.com/gwu-libraries/sfm-ui/issues/274

[...]

PR #305, opened 2016-06-03
Rm the old apache pid to fix the apache restart

Tanych:
The reasonable solution for the https://github.com/gwu-libraries/sfm-ui/issues/291


justinlittman:
Confirmed that this works.

Issue #285, opened 2016-05-24
UI should warn user if they are trying to run a filter with credentials already in use 


lwrubel:
Only one filterstream can run at a time with the same set of credentials.

kerchner:
This does not need to prevent multiple REST harvest Collections with the same credentials (vs. filter harvest Collections).   We'd have to determine whether Twitter's API allows simultaneous use of filter and REST API with the same credentials.

kerchner:
Allow credential to show up in the dropdown (selection of credentials), but don't allow the user to turn on collection of the stream.   Present a message stating that you can't turn on the filter because the credential is already in use by another filter (specify which filter, if possible).

Tanych:
The message notification would be like this:
The credential twitter-victor(_credential name linked to the credential_) is using by twitter search: twitter filter 1(_twitter search name linked to the collection details page_),  twitter filter 2,  twitter filter 3. Please stop them first, if you want to do the twitter filter.

So the credential is in use only by twitter filter, how about the credential is using by twitter search or twitter sample? Do I also need to forbid turning on if someone adds a twitter filter?

PR #312, opened 2016-06-09
Adding warning for twitter filter and sample

Tanych:
only allowed one collection of twitter filter and twitter sample to share the same credentials.
Filter one  --> on
Filter two --> can't turn on

Sample one  --> on
Sample two --> can't turn on

Refs: https://github.com/gwu-libraries/sfm-ui/issues/285

### DIFF
--- a/sfm/ui/views.py
+++ b/sfm/ui/views.py
@@ -75,7 +75,6 @@ class CollectionSetCreateView(LoginRequiredMixin, SuccessMessageMixin, CreateVie
         kwargs['request'] = self.request
         return kwargs
 
-
     def get_success_url(self):
         return reverse('collection_set_detail', args=(self.object.pk,))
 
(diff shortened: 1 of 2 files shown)


sfm/ui/views.py
```diff
@@ -142,6 +142,18 @@ def get_context_data(self, **kwargs):
             seed_warning_message = "At least 1 active seed must be added before harvesting can be turned on."
         context["seed_error_message"] = seed_error_message
         context["seed_warning_message"] = seed_warning_message
+
+        # Adding credential active message
+        context["credential_used_col_list"] = None
+        credential_warning_message = None
+        if self.object.harvest_type in [Collection.TWITTER_FILTER, Collection.TWITTER_SAMPLE]:
+            context["credential_used_col_list"] = Collection.objects.filter(credential=self.object.credential.pk,
```
justinlittman: This list might have 0 or 1 entries, correct?  If so, then don't pass a list, pass None or the single entry.  This will remove the need for the list iteration in the template.

Tanych:
@justinlittman  At first, I supposed to take twitter searches into consideration. There might be lots of possibilities. Right now, it's just the 0 or 1 issue.

justinlittman:
For the purposes of less complex code, please assume will be 0 or 1.

Tanych:
@justinlittman  I will fix it.

Issue #202, opened 2016-04-19
Update rabbitmq hash in docker-compose files to match sfm-docker prod.docker-compose.yml


kerchner:
In 0.5.0, the rabbitmq image hash in sfm-docker/prod.docker-compose.yml was updated to a more recent image.   However, we should also update the same in example.docker-compose.yml files across the various repos to match.

Tanych:
@kerchner If we update the `example.docker-compose.yml`, we also need to update `sfm-*-havester`.

kerchner:
Merged via multiple PRs above.

PR #363, opened 2016-07-07
keep match the rabbit image hash

Tanych:
Update the rabbit image hash
refs https://github.com/gwu-libraries/sfm-ui/issues/202


justinlittman:
@Tanych For trivial fixes like this, no need for pull request.  You can commit directly.

Issue #358, opened 2016-07-05
Tumblr Harvester / Exporter / etc.


kerchner:

justinlittman:
Notes on code review:
1.  The tumblr harvester can be made even simpler by getting rid of Oauth altogether.  If my reading to the tumblr api is correct, then we just append the consumer key as a parameter.  For example: https://api.tumblr.com/v2/blog/justinlittman.tumblr.com/posts?api_key=Fki0Q9w9QcW95yy66RtFCni14QpM0pjuHbDWMrZ9aPXcsthVQq
2. Number of retries, delays, etc. in tumblrarc should be configurable rather than hardcoded.
3. I'd suggest removing the dependency on twarc.
4. Please document the parameters for [blog_posts()](https://github.com/gwu-libraries/sfm-tumblr-harvester/blob/master/tumblrarc.py#L101).
5. According to the API documentation, the maximum number of posts that can be retrieved at a time is 20, not 50.
6. Unclear why you're breaking [here](https://github.com/gwu-libraries/sfm-tumblr-harvester/blob/master/tumblrarc.py#L145) if not incremental.
7.  Not sure position is a reliable way to store the state for incremental harvests.  What is a post is deleted in between harvests?  It will throw off the incremental retrieval.
8.  If [post](https://github.com/gwu-libraries/sfm-tumblr-harvester/blob/master/tumblrarc.py#L170) isn't used, then get rid of it.
9. tumblrarch needs unit tests (suggest vcrpy), especially for incremental harvests.
10.  It seems like there are 2 ways of exiting the loop.  1. Hit the end of the while and 2. Return 0 posts.  If logic is correct, should only need one or the other, not both.

justinlittman:
On TumblrHarvester:
1.  Instead of calling it user_posts, call it blog_posts (since the unit of harvest selection is a blog, not a user).
2. At https://github.com/gwu-libraries/sfm-tumblr-harvester/blob/master/tumblr_harvester.py#L43, "harvesting blog", not "harvesting user".
3.  Are you sure you want to add max_last_post and last_post at https://github.com/gwu-libraries/sfm-tumblr-harvester/blob/master/tumblr_harvester.py#L59?  I'm not following the logic.
4.  Other types of posts (e.g., link posts) can have photos, not just photo posts.
5.  What about media in audio and video posts?

justinlittman:
On TumblrWarcIter:
[...]

PR #373, opened 2016-07-14
Refs #358 supporting tumblr harvester

Tanych:
The PR for issue https://github.com/gwu-libraries/sfm-ui/issues/358 includes following:
* Adding `tumblr_user_posts ` harvester type in collection.
* Harvesting tumblr user posts with given token `blogname`.
* Supporting connect `tumblr` social account to add tumblr `credentials`.


sfm/ui/auth.py
```diff
@@ -6,7 +6,7 @@
 from django.core.urlresolvers import reverse
 from django.contrib import messages
 
-from .forms import CredentialTwitterForm,CredentialWeiboForm
+from .forms import CredentialTwitterForm,CredentialWeiboForm,CredentialTumblrForm
```
justinlittman: Add spaces after commas.

Tanych:
@justinlittman if we use `UID` as the blog_names, we need to add the deduplicate both for `token` and `UID` when people use single `Add Seed` form. While `Bulk add seeds` has already added deduplicate. For the harvester of `tumblr_blog_post`, the `Bulk add seeds`  will directly add seeds to `uid` instead of `token`.

Issue #378, opened 2016-07-18
Clarify weibo instructions


justinlittman:
From Nich Worby:

```
We’ve tried one crawl of Weibo and goofed up a little. None of us were regular Weibo users
so we weren’t 100% what a crawl of our timeline would yield. We weren’t following anyone at 
the time of the crawl and, as a result, got an empty crawl back. Section 4 of the Weibo 
instructions could benefit from having a few sentences letting us know what a timeline crawl 
will yield and how to set one up so that you get useable results. 
```

kerchner:
For now, update blog post in place.

PR #383, opened 2016-07-21
Refs #378 add public weibo timeline description

Tanych:
Ref #378 , to make the demo description more clear.


justinlittman:
Looks like you have some extra commits in this PR.

Tanych:
@justinlittman , this PR includes fixing the Travis buildings and an external link that it's invalid, both are tiny change. I didn't separate them.

justinlittman:
So what I think Nich was asking for is a description of friends timeline.  That is, it would be helpful to add another section describing what harvest we do with the weibo harvester (i.e., a friends timeline) and how it works.  (For example, that you have to follow the people whose weibos you want collect.)

Ideally, we should have a section in the user docs for this, but in the meantime a small addition to the blog post will suffice for explaining it for weibo.

justinlittman:
Also, I've reviewed everything that is committed to this branch so far and it is good to go.

Tanych:
Ok, I would add the section including the `friend timeline`.

PR #395, opened 2016-08-07
Improved clarity of title and labeled X axis - fixes #354. Also cente…

### LINKED ISSUE #354: Reconsider home page viz - dates of tweets
I found it confusing that tweet histograms are by date collected rather than date created.  Let's consider either plotting by date created, changing the plot label to something like "Tweets collected in {time period}".  If the main objective is to show collecting activity level, then perhaps the latter would make sense.

kerchner:
…red buttons and fixed capitalization of viz title upon update

### DIFF
--- a/sfm/ui/templates/ui/home.html
+++ b/sfm/ui/templates/ui/home.html
@@ -51,18 +51,16 @@
             </div>
             <div class="panel-body">
                 {% for item in collection_set.stats_items %}
-                    <div class="col-md-4">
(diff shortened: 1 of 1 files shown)


lwrubel:
@kerchner if you wanted to avoid adding a div just for centering the buttons, you could add text-center to the class for the col-md-4 div. Aside from that, looks good.

kerchner:
@lwrubel Aha, I had tried adding that to the innermost div, but that didn't work; your suggestion to add it to the col-md-4 div does work.

PR #403, opened 2016-08-11
fixes #350. Edits to change log help text

### LINKED ISSUE #350: Change log help text/in-app help
Consider example text or prompts for Change Log field, especially for what might be appropriate using the Create form vs Update.

lwrubel:


lwrubel:
Approach based on this: https://coderwall.com/p/-2er7a/make-your-modelforms-aware-of-whether-they-re-part-of-a-createview-or-an-updateview but open to other ideas.

kerchner:
@lwrubel is this ready for review?

lwrubel:
@kerchner, ready for review, looks like I forgot to assign someone.

kerchner:
@lwrubel It looks like your intention is that for new model objects, the Change Note message would be "Further information about this addition" versus the current message.  While the "add new" message is displayed for new collection sets, the "modify existing" message is still showing when I add a new seed, bulk-add a new seed list, add a new collection, or add a new credential.

This is likely because when adding a new collection set, there's no pk until after it's saved, whereas it appears that the other model object types are each given a provisional pk (not sure how to explain this for the add new credential screen).

lwrubel:
@kerchner, looks like I just hadn't applied this to the other CreateViews, thanks for noticing that. Ready for review again.

kerchner:
@lwrubel Bulk Add Seeds still shows the update message rather than the create message.

lwrubel:
@kerchner thanks for catching that. Fixed and ready for review.

Issue #409, opened 2016-08-18
Additional harvest notifications


justinlittman:
With harvests messages working correctly (#346, #394), it is clear that the results of harvests need to be brought to attention of users, especially when there are messages.

For this ticket, 2 notifications will be added:
1.  For a collection, if the last harvest failed or has messages, those will be provided as alerts on the collection detail page.
2.  If a harvest fails or has messages, send an email to the members of the collection's group.

justinlittman:
For testing, the easiest way to create error messages is to perform a twitter user timeline harvest with invalid screen names.

The easiest way to create a failed harvest is to perform a harvest using invalid credentials.

lwrubel:
Closed with #410.

PR #410, opened 2016-08-18
refs #409. Adds harvest notifications via email and on collection det…

justinlittman:
…ail page.

### DIFF
--- a/sfm/message_consumer/sfm_ui_consumer.py
+++ b/sfm/message_consumer/sfm_ui_consumer.py
@@ -61,28 +61,28 @@ class SfmUiConsumer(BaseConsumer):
         harvest.save()
 
         # Update seeds based on tokens that have changed
-        for id, token in self.message.get("token_updates", {}).items():
+        for seed_id, token in self.message.get("token_updates", {}).items():
             # Try to find seed based on collection and uid.
             try:
-                seed = Seed.objects.get(seed_id=id)
+                seed = Seed.objects.get(seed_id=seed_id)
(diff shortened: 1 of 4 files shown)


sfm/message_consumer/sfm_ui_consumer.py
```diff
@@ -98,6 +98,55 @@ def _on_harvest_status_message(self):
...
+            if receiver_emails:
+                harvest_url = 'http://{}{}'.format(Site.objects.get_current().domain,
+                                                   reverse('harvest_detail', args=(harvest.id,)))
+
+                # Send Status mail
+                if settings.PERFORM_EMAILS:
+                    if harvest.status == Harvest.SUCCESS:
+                        mail_subject = u"SFM Harvest for {} completed successfully, but has messages".format(
+                            harvest.collection.name)
+                        mail_message = u"The harvest for {} ({}) completed successfully.".format(
```
lwrubel: @justinlittman is there a reason the message doesn't also say "but has messages"? When viewing the email, it would be useful to have the full info in the body.

justinlittman: @lwrubel What do you mean by "have the full info in the body." Do you mean add "but has messages" in the body or something more?

lwrubel: Yes, that's all. The "but has messages" part.

justinlittman:
@lwrubel Fixed and rebased.

Issue #414, opened 2016-08-23
collection detail size lable


Tanych:
The image size label in collection detail page currently  is `sizes`, sometimes it's  a little tricky to understand. using `Image sizes` instead.

PR #415, opened 2016-08-23
Ref #414 tweaks on image sizes

Tanych:
Ref #414


kerchner:
@Tanych have you considered binding some of the hard-coded `label` values for Image size  and web resources similar to https://github.com/gwu-libraries/sfm-ui/blob/9b73e9414bfdc16b09c80a9fe766de04213e91e4/sfm/ui/forms.py#L34-L38 ?

Tanych:
@kerchner  those labels are used for the collection update pages and the `sizes` having been modified in the label field.

Issue #404, opened 2016-08-15
Add ability to bulk-add seeds by UIDs


kerchner:
Along with this, we should validate:
- when adding by Twitter screennames, prevent the user from adding screennames that are entirely numeric (this currently throws an exception - see #355 
- when adding by Twitter UIDs, prevent the user from adding screennames that are _**not**_ entirely numeric (should double-check that entirely numeric screennames are disallowed)

Also verify how this works for other platforms besides Twitter (for example, Flickr usernames vs. nsid-s).

There are different ways we can design the UI, but I would suggest a single bulk list entry field, plus a radio button (prominent, perhaps above the list-entry field) where the user selects screennames vs. UIDs.  Suggest defaulting the radio button value to "None" so that the user must consciously select the value.  The value would be required.

Tanych:
I am not sure whether adding radio button can avoid the invalid input since the users can also input  entirely numeric when they choose `screennames`. What I think about this issue is that users added the bulk seeds (token or uids) in a single bulk list. We separate the uids with some patterns like twitter uids are entirely numbers and Flickr users seems like `1233@N001` .  We'd better discuss a little more in this issue.

kerchner:
Note that the scope of this issue is limited to Twitter.  Analogous enhancements to other platforms' seed management will be in separate, future tickets.

Issue #355, opened 2016-07-05
Bulk add seeds by UID results in error


kerchner:
Harvest returned: `'int' object has no attribute 'lstrip'`

This would seem to indicate the app doesn't currently allow bulk add by UID.  This ticket could be parlayed into an enhancement.

kerchner:
Ref #353

kerchner:
Add warning and cast as strings.   And/or allow user to specify whether bulk list is UIDs or account names.

kerchner:
This should be resolved by #404 as long as the validation piece described in #404 is implemented.

PR #427, opened 2016-09-08
Ref #404 reconsiderations for the bulk seeds

Tanych:
ref #404 #355

### DIFF
--- a/sfm/ui/forms.py
+++ b/sfm/ui/forms.py
@@ -619,10 +619,10 @@ class BulkSeedTwitterUserTimelineForm(BaseBulkSeedForm):
                 strtoken.append(clean_t)
         if seed_type == 'token' and numtoken:
             raise ValidationError(
-                'Screen name should not contain any entirely numeric seed! Invalid seeds: ' + ', '.join(numtoken) + '.')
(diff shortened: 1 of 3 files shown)


kerchner:
@Tanych  The bulk seeds add page works, but validations need to be implemented as per the first two bullet points in #404 .

Tanych:
@kerchner 
At first, I try to add some specify length rules for the `sceen_name` and `uid`, however, when I find the definiton of the two fields, I roll back what I was thinking.
* **id**: The integer representation of the unique identifier for this User. This number is greater than 53 bits and some programming languages may have difficulty/silent defects in interpreting it. Using a signed 64 bit integer for storing this identifier is safe. Use id_str for fetching the identifier to stay on the safe side. See Twitter IDs, JSON and Snowflake.
* **screen_name**: The screen name, handle, or alias that this user identifies themselves with. screen_names are unique but subject to change. Use id_str as a user identifier whenever possible. Typically a maximum of 15 characters long, but some historical accounts may exist with longer names.

I can add the two bullet points in #404, but it can't avoid the test cases problems.
Refs:
https://dev.twitter.com/overview/api/users

kerchner:
Fixes #355, Fixes #404

Issue #420, opened 2016-09-01
Twitter filter harvests stop collecting


justinlittman:
On at least 2 occasions, twitter filter streams running in production have stopped collecting.  They are still running, but spinning on errors (see below).

justinlittman:
Here's the relevant log from a failure in production:

```
2016-09-04 01:00:43,107: __main__ --> Processed 20700 tweets
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/SocketServer.py", line 599, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/local/lib/python2.7/SocketServer.py", line 334, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/local/lib/python2.7/site-packages/warcprox/warcprox.py", line 158, in __init__
    warcprox.mitmproxy.MitmProxyHandler.__init__(self, request, client_address, server)
  File "/usr/local/lib/python2.7/site-packages/warcprox/mitmproxy.py", line 25, in __init__
    http_server.BaseHTTPRequestHandler.__init__(self, request, client_address, server)
  File "/usr/local/lib/python2.7/SocketServer.py", line 655, in __init__
    self.handle()
  File "/usr/local/lib/python2.7/BaseHTTPServer.py", line 340, in handle
    self.handle_one_request()
  File "/usr/local/lib/python2.7/BaseHTTPServer.py", line 328, in handle_one_request
    method()
  File "/usr/local/lib/python2.7/site-packages/warcprox/mitmproxy.py", line 96, in do_CONNECT
    self.handle_one_request()
  File "/usr/local/lib/python2.7/BaseHTTPServer.py", line 328, in handle_one_request
    method()
  File "/usr/local/lib/python2.7/site-packages/warcprox/mitmproxy.py", line 131, in do_COMMAND
    self._proxy_request()
  File "/usr/local/lib/python2.7/site-packages/warcprox/warcprox.py", line 210, in _proxy_request
    buf = h.read(8192)
  File "/usr/local/lib/python2.7/httplib.py", line 588, in read
    return self._read_chunked(amt)
  File "/usr/local/lib/python2.7/httplib.py", line 630, in _read_chunked
    line = self.fp.readline(_MAXLINE + 1)
  File "/usr/local/lib/python2.7/site-packages/warcprox/warcprox.py", line 125, in readline
    hunk = self.fp.readline(size)
[...]

Issue #320, opened 2016-06-13
Reconsider Twitter search


justinlittman:
When running a Twitter search for the Pulse shootings, but the following was observed:
* Search ran until it hit a rate limit, paused, and repeated.  The search never ended.
* All of this was recorded to a single WARC, which was never finished.
* Other harvest requests were blocked waiting for the search  to end.

Consider:
* Only running search backwards (which should be finite).
* Terminating search after a specific amount of time or number of tweets.

Tanych:
I realize the same problem one week ago when I am trying to search NBA playoff. It needs to wait a long time to complete the search.

justinlittman:
I've reviewed the Twitter search API docs.  The reality is that a search can take a long time given the number of possible tweets and rate limits.

I think the solution is to run multiple Twitter REST harvesters so that searches can run without holding up other harvest requests.  Also, suggest (in docs) against running long searches with too short a frequency.

I'm going to do some experiments with running multiple harvesters.

justinlittman:
Also need to consider whether warcprox is holding open a single WARC file for an extended period (which can be risky) or creating multiple WARCs.

kerchner:
Implemented by:
- https://github.com/gwu-libraries/warcprox/commit/b015e4ae804d32371bb0492d63b11c0504c4f319
- documentation updates to come

kerchner:
Fixed in #444 and in the above change to warcprox.

PR #444, opened 2016-09-16
refs #420. Turns off streaming harvests that return failure.

justinlittman:


kerchner:
@kerchner (note to self) speak to @justinlittman first re: environment setup for this ticket.

kerchner:
Also test #320 - observe on a Twitter search that WARCs are written every 30 minutes (or as set)

kerchner:
@justinlittman Add a note with the entry in the change log generated when the system deactivates the harvest to note the reason.

kerchner:
- [X] Observed that Twitter search closes out WARCs at 30 minutes
- [X] Observed that Twitter filter with bad credentials is turned off by system

Issue #428, opened 2016-09-09
Monitor free space


justinlittman:
Well, we got bitten again running out of free space on /sfm-data (this time on sfm-test).  This has been the single largest cause of our failures.  While in theory monitoring disk space should be part of the infrastructure, as a practical matter SFM should monitor and provide warnings.

SFM will check disk space on /sfm-data and /sfm-processing at configurable intervals. When disk space falls below a configurable level, it will send an email to the configured admin. It will not send another email until disk space goes above the threshold.

It is recommended to implement this as a scheduled job in SFM UI.  See the user harvest emails code for an example of performing tasks on a schedule.

justinlittman:
Also worth considering is displaying free space on the home page of admins.

justinlittman:
Before working on this, let's finalize the requirements.

justinlittman:
If you logged in as an admin, then on the home page the amount of free space on /sfm-data and /sfm-processing will be displayed.  Also, the amount of free space will be included in the harvest status email message for admins.

Tanych:
It might be something like the [cAdvisor](https://github.com/google/cadvisor), which give the usage of the containers. It got more advanced, for this issue, we need a simple one to monitor the space of /sfm-data and /sfm-processing and then sending email notifications. However, the sfm-data  and sfm-processing get exit when setup, the cAdivosr has no information about the exit containers.  We can talk a little more about the issue.

I tried some pieces for the CAvisor, it gives me some graph of the usages, the following graph is the **twitter-harvester container**:
![cup_mem](https://cloud.githubusercontent.com/assets/3956411/19127560/79612066-8b0d-11e6-8110-c6851b4fd025.png)

justinlittman:
@Tanych cAdvisor is overkill for this ticket, but it looks real interesting.  Can you move the work on cAdvisor to #470?

For this ticket, we're looking for something much simpler -- the python equivalent of `df -h`.

Tanych:
@justinlittman Yeap, for this ticket, we only need to get a very simpler solution to do that. Just check the space. I just did some researches for the possible use in the future.  I will focus on solving this tickets first. The cAdvisor would take consideration later.

Tanych:
Some preview works for the emails and admin index:
* Emails Notification

![emails_notifaction](https://cloud.githubusercontent.com/assets/3956411/19444155/4b9899a8-945d-11e6-8b08-be0e15ca7846.png)

* Admin home
[...]

PR #487, opened 2016-10-18
Ref # 428 Monitor free space

Tanych:
Ref https://github.com/gwu-libraries/sfm-ui/issues/428

### DIFF
--- a/sfm/ui/notifications.py
+++ b/sfm/ui/notifications.py
@@ -2,6 +2,7 @@ import logging
 import datetime
 from collections import OrderedDict
 from smtplib import SMTPException
+from subprocess import check_output, CalledProcessError
 
 from django.template.loader import get_template
 from django.template import Context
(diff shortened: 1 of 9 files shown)


sfm/ui/notifications.py
```diff
@@ -11,13 +12,166 @@
...
+        :param threshold: the free space threshold.
+        :return:
+        """
+        # deal with the empty string
+        if not volume_dir:
+            volume_dir = 'None'
+        if not threshold:
+            threshold = '10GB'
+
+        self.space_msg_cache = {'volume_id': volume_dir, 'threshold': threshold, 'bar_color': 'progress-bar-success'}
```
justinlittman: You don't need to make the change, but a slightly more pythonic way of handling the empty string would be:
```
self.space_msg_cache = {'volume_id': volume_dir or 'None', 'threshold': threshold or '10GB', 'bar_color': 'progress-bar-success'}
```

Issue #303, opened 2016-06-03
Need mechanism for monitoring harvesters/exporters


justinlittman:
Both whether they are up and what they are working on and how far they've gotten on what they're working on.  Other than peeking at the logs.

kerchner:
Also consider monitoring components such as UI app, message queue, etc. - health check.

justinlittman:
Branches for this ticket:
* https://github.com/gwu-libraries/sfm-ui/tree/t303-monitor
* https://github.com/gwu-libraries/sfm-utils/tree/sfm_303-monitor
* https://github.com/gwu-libraries/sfm-docker/tree/sfm_303-monitor

justinlittman:
I scoped this ticket down to functionality that could be added with a low effort level, but provided a reasonable first cut at monitoring.

justinlittman:
To create test environment:
* Add `RABBITMQ_MANAGEMENT_PORT` to environment for ui. See https://github.com/gwu-libraries/sfm-docker/compare/sfm_303-monitor for changes to be made to docker-compose.yml file.
* Run all containers in dev:
```
UI_REQS=dev
TWITTER_REQS=dev
FLICKR_REQS=dev
TUMBLR_REQS=dev
WEIBO_REQS=dev
WEB_REQS=dev
```
Also, make sure that code is linked in.  For example:
```
        volumes:
            - "../sfm-twitter-harvester:/opt/sfm-twitter-harvester"
            - "../sfm-utils:/opt/sfm-utils"
            - "../warcprox:/opt/warcprox"
```

To test:
1. After startup, click Monitor on the header. Notice that there is no status or queued requests.
2. Create some harvests and exports.  Notice that there statuses appear in Monitor for harvesters and exporters.  Notice that only the most recent harvest/export that is handled is listed.  Note unless your timing is really good, you won't see any queued requests.
3.  Kill a harvester and exporter.  Now create some harvests for the killed harvester and exports for the killed exporter.  Notice that queued requests are now listed.
4.  Kill UI and bring back up.  Monitors should still appear properly, though it will list new harvester/exporter instances. (One is the old instance, the other is the new instance.)

justinlittman:
Here are the PRs for this ticket:
* https://github.com/gwu-libraries/sfm-ui/pull/499
* https://github.com/gwu-libraries/sfm-utils/pull/22
* https://github.com/gwu-libraries/sfm-docker/pull/9

justinlittman:
[...]

Issue #362, opened 2016-07-07
Enforce authorizations consistently in UI


lwrubel:
See description of issue in: https://docs.google.com/document/d/1eCUsHB2x_gGvxItcfGPLfphOm9C8W7gC07Udxs3nC4k/edit#

lwrubel:
Recommend including in 1.5

justinlittman:
This ticket will consolidate #571 and #493, which will be closed.

justinlittman:
Requirements are as follows:
1.  A user will only be able to view and edit collection sets, collection, seeds, harvests, and exports for groups with which she is affiliated.  This includes directly typing in a URL to get to a page.  On Monitor page (or any similar page), links will only be provided by items that the user can access.
2.  Staff can view all, but only edit those associated with groups with which they are associated.  A staff may request an export from all collections.
3.  Superusers can view and edit all.
4.  A user or staff can only view the credential details for her own credential.  This also includes the credential history.
5.  Superusers can view all credentials.

kerchner:
0.  There are three levels of users:   User, Staff, Superuser
6.  Superusers can access the Django admin page (staff and users cannot)

Just to acknowledge, this is an expansion of the scope of the ticket, which was originally just to enforce the current scheme.

I assume with the scope as above that granting users higher roles would still occur via the django admin page.

With regard to #450, the implication seems to be that when implemented, #450 would authenticate users but groups/roles would be assigned within the SFM app, rather than pulling from LDAP group membership. (I would advocate for the former as being easier to manage).

justinlittman:
This is probably part of the implementation approach:  https://docs.djangoproject.com/en/1.8/topics/auth/default/#django.contrib.auth.decorators.user_passes_test

Tanych:
## User Level
create two users
* user1(group: user1) 
* user2(group:user2)

Building collection sets, collection and check whether they can view each other. 
### View and Edit Result
✓: mean can  X: mean can't

| collection set | collection | seeds | harvester | exporter | credential |
|----------------|------------|-------|-----------|----------|------------|
| X              | X          | X     | X         | X        | X          |

## Staff Level
create one staff and associated `user2`
* staff1(group:user2)
### Staff View Result
* Group user1

| collection set | collection | seeds | harvester | exporter | credential |
|----------------|------------|-------|-----------|----------|------------|
| ✓              | ✓         |✓     | ✓         | **X**        |  **X**         |

* Group user2

[...]

PR #499, opened 2016-10-26
refs #303. Adds monitoring page and monitoring docs.

justinlittman:


lwrubel:
@justinlittman: I tried changing the port that Rabbit MQ runs on (so I could use a port that's open on my WRLC vm). I see that common.py has the following settings, I don't think they should have the "SFM_" at the beginning of the setting name. And I can't find SFM_RABBITMQ_HOST in example.env. Where is that coming from?

```
RABBITMQ_HOST = env.get('SFM_RABBITMQ_HOST')
RABBITMQ_USER = env.get('SFM_RABBITMQ_USER')
RABBITMQ_PASSWORD = env.get('SFM_RABBITMQ_PASSWORD')
RABBITMQ_MANAGEMENT_PORT = env.get('SFM_RABBITMQ_MANAGEMENT_PORT', "15672")
```

justinlittman:
I confused myself for a minute.  Some of this is overridden by
docker_settings.py, which contains:
RABBITMQ_HOST = "mq"
RABBITMQ_USER = env.get('RABBITMQ_USER')
RABBITMQ_PASSWORD = env.get('RABBITMQ_PASSWORD')

lwrubel:
The monitor page shows all harvests, regardless of whose collections they're for. Is that intentional? I can sort of understand the scenario, that you want to see if your harvests are stuck because of someone else's harvests. But then you can link right into that collection and edit it, even though it's run by someone else.  In the meantime, perhaps only link harvests from collection collection sets that are in the group the user belongs to? I'm open to this adjustment being an issue in 1.4.  #362 would address this more fully, will recommend that for 1.5. 

The page is also accessible without login, which seems odd (although links into harvest pages require login, so you don't get too far).

PR #515, opened 2016-11-04
Fixes #433 - Adds user images and banner to web harvest

### LINKED ISSUE #433: Consider including URLs in user profile (images/links) in web resources harvest
Maybe just for Twitter user timeline collection?  This would be extremely valuable to the Program on Extremism use case, or any other researchers studying user profiles.  Probably would be easy to implement; just expand the scope of URLs that are harvested to include URLs within the user block of the tweet JSON.

adityadharne:
Ref: https://github.com/gwu-libraries/sfm-twitter-harvester/pull/22


justinlittman:
You also want the profile_banner_url.  This is header photo that appears on profile pages.

Also, the changes in sfm twitter harvester need unit tests.

Issue #387, opened 2016-07-26
Add instructions for upgrading existing instance to new versions of docker containers


kerchner:
This should be fairly simple (edit docker-compose, re-compose, etc.) but want to make sure we don't forget to add this before we release the first post-1.0 version.

justinlittman:
See #302.  This will need some testing and probable enhancements of harvesters and exporters to make sure upgrade is graceful.

kerchner:
👍

PR #522, opened 2016-11-11
refs #387. Adds upgrade instructions.

justinlittman:


justinlittman:
@lwrubel Can you give this a quick read-through?  I'll be testing as part of our next upgrade.

PR #548, opened 2016-11-18
Changed the introduction to credentials

yonahbg:
Included more detail about why each user needs their own credentials. First element for t544 addressed.


lwrubel:
Nice improvements, @yonahbg. I made a small tweak to the prior programming suggestions. This is live now.

PR #553, opened 2016-11-23
Added post about jq notebook.

justinlittman:


lwrubel:
Either in the post or the notebook itself, you might link to Matthew Lincoln's nice jq tutorial, which has some other nice recipes and explanation. http://programminghistorian.org/lessons/json-and-jq.html. Alternatively, I could update the Getting Started notebook to point to it.

justinlittman:
@lwrubel Matthew Lincoln's jq tutorial is linked [here](http://sfm.readthedocs.io/en/latest/processing.html#using-jq-to-process-json).  However, adding it to the Getting Started notebook wouldn't hurt either.

Issue #513, opened 2016-11-03
Provide guidance on server sizing


justinlittman:
As suggested by @ianmilligan1:
```
 Some guidance on EC2 instance size might help – hard to find the sweet spot w/r/t how much memory/CPU to run.
```
https://twitter.com/ianmilligan1/status/794171334192406528

PR #572, opened 2016-12-05
refs #513. Adds guidance on server sizing.

justinlittman:


kerchner:
@justinlittman Did you intend to leave the production server type blank, in the table?

justinlittman:
Yes, since it is not AWS.

Issue #501, opened 2016-10-27
Document storage


justinlittman:
As recommended by @dchud:
- Explain linked vs. Docker storage
- Storage directory structure
- How to copy out of Docker storage
- How to migrate between Docker storage and linked storage.

lwrubel:
This is a documentation ticket.

justinlittman:
Discussion with @ianmilligan1 about experiences setting up SFM illustrated that these docs would be helpful.  Make sure to reference the docs in example.env.

justinlittman:
@chrisprom also asked for explanation of linked vs docker storage.

PR #576, opened 2016-12-06
refs #501. Storage docs

justinlittman:


kerchner:
@justinlittman should the copy's target location in this example
https://github.com/gwu-libraries/sfm-ui/pull/576/files#diff-66f1a25d8d482134be470912443ee224R110
be `/` or `/sfm-data` ?

justinlittman:
Yes.  I re-tested to verify.

Issue #530, opened 2016-11-15
Test sparse Twitter filters


justinlittman:
This came out of testing by Maria at IA.

The monitoring available for Twitter filters that collect sparse tweets may be confusing.

To test:
1. Create a Twitter filter for something really obscure.
2. Look at the harvest list.
3. Wait an hour.
4. Look at the harvest list.
5. Tweet something that should be captured by the Twitter filter.
6. Wait a half hour.
7. Look at the harvest list.
8. Wait an hour.
9. Look at the harvest list.

Expected result:
Each time you look at the harvest list it should convey useful and clear information.

justinlittman:
This is done.

Issue #472, opened 2016-10-06
Prevent harvesting congestion


justinlittman:
This is based on testing by Internet Archive.

Huge backlogs can occur when a user creates a number of long-running harvests (broad Twitter searches, Twitter user timelines with a large number of seeds) and those harvests share a small number (or one) set of credentials.

This problem is not resolved by adding more harvesters, as running harvests with the same set of credentials hits rate limits more quickly and clogs the harvester.

Further, as the backlog grows there may be multiple harvest requests for a single collection in the queue.

Some possible approaches:
- Limit the number of collections a set of user credentials can be used with (or warn the user when a set of credentials is used repeatedly).
- Make queues that are specific to a set of credentials.
- Move the scheduler into the harvester.  This way, the scheduler can be aware of currently running harvests.

justinlittman:
Here's a simple approach:
1. Change the scheduler so that it will not send a new harvest request if the last harvest has not completed.
2. If a harvest is skipped, send an email with warning.  May be due to too short a frequency, problem with last harvest, or not enough harvesters.
3. Provide the ability to void a harvest. That is, mark it as a failure so that a new harvest can be scheduled.
4. Warn users when they select credentials that are already in use by another collection.  (Twitter only?)

PR #582, opened 2016-12-08
refs #472. Adds support for voiding harvests. refs #530. Changes harv…

justinlittman:
…est stats message when harvest is running.

### DIFF
--- a/docs/troubleshooting.rst
+++ b/docs/troubleshooting.rst
@@ -18,6 +18,18 @@
  Specific problems
 -------------------
 
+Skipped harvests
+^^^^^^^^^^^^^^^^
+A new harvest will not be requested if the previous harvest has not completed. Instead, a harvest record will be created
+with the status of skipped. Some of the reasons that this might happen include:
+* Harvests are scheduled too closely together, such that the previous harvest cannot complete before the new harvest is
+  requested.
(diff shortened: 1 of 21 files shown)


justinlittman:
For testing the credential use warnings:
1.  Create two Twitter credentials.
2.  Use one of those Twitter credentials to create a user timeline collection.
3.  Go to create user timeline page.  Observe results when selecting different credentials.  For the credential which is in use by another collection, should receive warning that credential is in use.  For the credential which is not in use by another collection, should receive message that not in use.
4.  Also test same for editing an existing user timeline collection.
5.  Create a Flickr credential.
6.  Go to create Flickr user page.  Observe that do not receive warnings when selecting credentials.

For testing that additional harvest requests are not added to the queue when previous requests have not finished:
1.  Turn off a harvester.
2.  Create a collection for that harvest type scheduled for every 5 minutes.  Turn it on.
3.  Wait.
4.  A harvest record should be created whose status is skipped.
5.  Admins and members of the collection's group should receive emails notifying that the harvest was skipped.
6.  Go to the harvest detail page of the harvest that is stuck and void it.
7.  Turn harvester back on.
8.  Wait.
9.  Harvests should resume.

Also, the free space emails should be re-tested.  (To do this, just set the thresholds higher than the amount of available free space.)

adityadharne:
Tested successfully

Issue #479, opened 2016-10-14
Delete


justinlittman:
For various reasons, I find myself needing to delete (primarily in testing/dev environments).  It would be helpful if SFM UI supported the following forms of delete:
- Delete collection set and all data on disk.
- Delete collection and all data on disk.
- Delete all harvests for a collection and all data on disk.

justinlittman:
Depends on #532

kerchner:
Fixed in #589

PR #589, opened 2016-12-09
refs #479. Adds support for deleting collections, etc.

justinlittman:


justinlittman:
Steps for testing:
1.  Create collection sets, collections, and exports.
2.  Check that the files exist in /sfm-data.
3.  Log into Admin and delete them.
4.  Check that the files no longer exist.

Test with Apache and runserver.

adityadharne:
@justinlittman I am getting a `[Errno 30] Read-only file system` error. Should I change permissions for the directory?

justinlittman:
In your docker-compose.yml make sure that the volumes that are linked into the ui container are not marked "ro".

```
        volumes_from:
            - data
            - processingdata
```

adityadharne:
Delete works. Merging

Issue #377, opened 2016-07-18
Display seed changes on collection page


justinlittman:
To show when seeds were added/changed to the collection.

justinlittman:
Given django-simple-history, what is needed is a union of the collection historical objects and the seed historical objects.  However, django querying doesn't provide a good way to do this.

One approach is to merge the diffs.  The procedure used https://github.com/gwu-libraries/sfm-ui/blob/t123-manifest/sfm/ui/export.py#L89 can also be used for this ticket.

justinlittman:
This was easy, given #123 .

Issue #123, opened 2016-01-11
Create initial export manifest


kerchner:
Followon to #67.  Create initial infrastructure to generate an export manifest, to include basic metadata generated from WARCs and/or from app logs.

kerchner:
Discuss with broader project team to assess current work and current thinking on export manifest.

lwrubel:
Depends on work by Chris Prom exploring needs and use cases, and consideration by archivists.

justinlittman:
Depends on #532 to allow writing to /sfm-data from sfm ui.

justinlittman:
TODO:  Add to documentation.

justinlittman:
Also, need to remove RO from /sfm-processing in sfm ui container.

justinlittman:
Depends on #532.

lwrubel:
Agreed to go with a basic export manifest of collection / collection set metadata, in the form of a README as created for the elections dataset, and request feedback before further development.

justinlittman:
https://github.com/gwu-libraries/sfm-docker/commit/844e6356919aff3bc93d5d40a9e464841e67e178 removed RO from /sfm-processing.

kerchner:
PR is https://github.com/gwu-libraries/sfm-ui/pull/591

PR #591, opened 2016-12-09
refs #123. Adds README to exports and management command to generate …

justinlittman:
…READMEs. refs #377. Displays seed change log on collection page.


justinlittman:
Suggestions for testing README:
1.  Create an export.
2.  Observe that a README.txt is included in the export.  Check the content of the README.

Suggestions for testing change log:
1.  Go to a collection detail page which has seeds.
2.  Observe that change logs includes changes of collection and seeds.  What is being changed is indicated by Change to column.

lwrubel:
Comments on README:
* README is showing, "Format: export.export_format" when it should be "CSV" (or other format selected). 
* This is minor, could be addressed later, but the harvest options are in a different order than the UI. 
* Date created is not same timezone: "Created on Dec. 15, 2016, 8:32 a.m." when it was created at 1:32 p.m.

lwrubel:
I'm getting an error when trying an export where the collection has more than one seed. 
```
Exception Type:	TypeError
Exception Value:	
'str' object is not callable
Exception Location:	/opt/sfm-ui/sfm/ui/forms.py in label_from_instance, line 793
```

for line:

`{% for choice in field.field.choices %}`

lwrubel:
{{ item.collection.collection_set.name }} is displaying instead of the collection_set name in breadcrumbs at for seed change_log pages. Django doesn't seem to like line breaks in template variables.

justinlittman:
@lwrubel Fixed problems that you reported.

PR #596, opened 2016-12-15
Refs 544 Adds detailed instructions for getting credentials

yonahbg:
Not sure if the refs work properly, they need to be checked.


justinlittman:
@yonahbg Is this ready for review?

yonahbg:
Yes

yonahbg:
Actually, there are a few changes to be made based on the comments on t#544

Issue #379, opened 2016-07-18
Re-apply for advanced API access to Sina Weibo


kerchner:

Tanych:
## First trial was rejected! 
Reason:
```
搜索话题高级接口 2016-07-19 申请驳回 应用使用人数不符合接口申请条件
```
>The number of the users in the apps doesn't meet the application requirement.

It didn't tell me how many users the app should be. My test application only 2 users.

kerchner:
How do they measure how many users the app has?   Does this mean how many simultaneous users the app can handle?

Tanych:
@kerchner  the app has a management UI can show the statistics.

kerchner:
@Tanych when you have a chance, you can update with your latest findings

Tanych:
@kerchner 
Currently, the Weibo official has passed the first step screen process for the app `socialfeedmanager`. The next step is app store verification. It might take days. I already get some users to try the authorization work.

kerchner:
@Tanych Great, thanks for the update!

Tanych:
@kerchner 
The two advanced APIs application  has submitted, waiting for the response.

![image](https://cloud.githubusercontent.com/assets/3956411/21668731/95e42d62-d33f-11e6-92d4-1b8f73260509.png)

kerchner:
@Tanych can provide credentials w/API tokens

Tanych:
@kerchner 
I have wrote the work process for you to test the Weibo Search PRs since it might make you confused when lots of things involved.

## Guide to test the Weibo Search
### Weibo search PR related
* sfm-ui: https://github.com/gwu-libraries/sfm-ui/pull/624
* sfm-docker: https://github.com/gwu-libraries/sfm-docker/pull/17
* sfm-weibo-harvester: https://github.com/gwu-libraries/sfm-weibo-harvester/pull/27

### Weibo Advanced Doc PR:
* sfm-ui post: https://github.com/gwu-libraries/sfm-ui/pull/634

## Weibo Search Process
### Weibo Search Turn Off
* keep  the Weibo search option as `OFF` in  sfm-docker [.env](https://github.com/gwu-libraries/sfm-docker/blob/t630-weibo_hide/example.env#L79)
* Check whether the collection set has Weibo search

### Weibo Search Turn on
* Turn on the Weibo search option on sfm-docker [.env](https://github.com/gwu-libraries/sfm-docker/blob/t630-weibo_hide/example.env#L79)
[...]

PR #603, opened 2016-12-19
Ref #379 add weibo share button

Tanych:
add weibo share button on the home page to make pass the weibo screen process
* temporary add the button on the twitter widget 
* position can be re-design when get pass


Tanych:
@lwrubel @justinlittman @kerchner 
I will temporary merge this PR since the timezone diff in China. Thus, the Weibo official can do their screen work.

Issue #613, opened 2016-12-22
exporter readme unicode issue


Tanych:
* Test case
create a exporter based on any export whoes creadiental name is unicode, or like this:
```
Change to weibo_export (collection) on Dec. 22, 2016, 3:59:45 p.m. EST by victor:
* credential: "blank" changed to "百里紫砂's weibo credential"
* harvest_type: "blank" changed to "weibo_timeline"
* name: "blank" changed to "weibo_export"
* collection_set: "blank" changed to "weibo_Testing"
* is_active: "blank" changed to "False"
* schedule_minutes: "blank" changed to "10080"
* harvest_options: "blank" changed to "{"image_sizes": [], "incremental": true, "web_resources": false}"
```
* Error

```python 
Traceback (most recent call last):
  File "/opt/sfm-ui/sfm/manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 346, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/base.py", line 394, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/base.py", line 445, in execute
    output = self.handle(*args, **options)
  File "/opt/sfm-ui/sfm/message_consumer/management/commands/startconsumer.py", line 20, in handle
    consumer.run()
  File "/usr/local/lib/python2.7/site-packages/kombu/mixins.py", line 177, in run
    for _ in self.consume(limit=None):  # pragma: no cover
  File "/usr/local/lib/python2.7/site-packages/kombu/mixins.py", line 199, in consume
    conn.drain_events(timeout=safety_interval)
  File "/usr/local/lib/python2.7/site-packages/librabbitmq/__init__.py", line 220, in drain_events
    self._basic_recv(timeout)
  File "/usr/local/lib/python2.7/site-packages/kombu/messaging.py", line 598, in _receive_callback
    return on_m(message) if on_m else self.receive(decoded, message)
  File "/usr/local/lib/python2.7/site-packages/kombu/messaging.py", line 564, in receive
    [callback(body, message) for callback in callbacks]
[...]

PR #614, opened 2016-12-22
Ref #613 make file write support unicode

Tanych:
Ref https://github.com/gwu-libraries/sfm-ui/issues/613


Tanych:
@justinlittman I think it'd better merge this PR and re-tag 1.4 in sfm-ui since the export `readme.txt` issue.

Issue #257, opened 2016-05-16
Support branding of UI


justinlittman:

kerchner:
Also include contact info as part of branding info.

kerchner:
Summer 2016 implementation should be min.

kerchner:
- include a contact address for assistance.

lwrubel:
@lwrubel suggest further detail on this ticket

kerchner:
@lwrubel will propose a scope for this ticket

lwrubel:
Assignment for work to implement branding TBD, likely @Tanych or @adityadharne

lwrubel:
Contact email address. 
Institution name
Institution URL
Institution logo

Consider .env file as location for these elements.

lwrubel:
Create a new, separate setting in .env for contact email (different from admin email). After discussion, put institutional info in the footer.

justinlittman:
@lwrubel Does the institutional info appear as you'd like?
![screen shot 2017-01-18 at 8 34 11 am](https://cloud.githubusercontent.com/assets/588335/22066039/ffaadbec-dd58-11e6-8c0d-10cdfaa43bc2.png)

justinlittman:
This is tested and good to go once @lwrubel approves institutional info.

justinlittman:
Pull requests:
* https://github.com/gwu-libraries/sfm-docker/pull/14
* https://github.com/gwu-libraries/sfm-ui/pull/617

These will need to be squashed/rebased before merge.

lwrubel:
Per discussion, make institution name and link configurable, remove copyright reference. Everything else looks good.

Tanych:
@justinlittman  
make some tweaks on PRs 
* https://github.com/gwu-libraries/sfm-docker/pull/14
* https://github.com/gwu-libraries/sfm-ui/pull/617

The default value is `gw libraries`  with the corresponding links, it looks like:

![new_footer](https://cloud.githubusercontent.com/assets/3956411/22434678/ff498d60-e6ea-11e6-8e5b-9777c4008b90.png)
Please take a look at it.

justinlittman:
Also, https://github.com/gwu-libraries/sfm-docker/commit/6b2e8ea78eca819a94c9bb97873b35a471b14eb6

PR #617, opened 2017-01-03
Ref #257 UI branding

Tanych:
Ref https://github.com/gwu-libraries/sfm-ui/issues/257
* re-design the position of the footer compatible for mobile
* adding contact email info 
* adding gw libraries name
* adding copyright info
* adding links for http://library.gwu.edu

### DIFF
--- a/sfm/sfm/settings/common.py
+++ b/sfm/sfm/settings/common.py
@@ -166,6 +166,12 @@ REST_FRAMEWORK = {
     'DEFAULT_FILTER_BACKENDS': ('rest_framework.filters.DjangoFilterBackend',)
 }
 
+# Institution name, such as 'GW Libraries'
+INSTITUTION_NAME = env.get('SFM_INSTITUTION_NAME')
+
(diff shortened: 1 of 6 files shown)


sfm/sfm/settings/common.py
```diff
@@ -193,6 +193,9 @@
 EMAIL_HOST_PASSWORD = env.get('SFM_EMAIL_PASSWORD')
 EMAIL_USE_TLS = True
 
+# contact email
+CONTACT_EMAIL = env.get('SFM_CONTACT_EMAIL', 'sfm@gwu.edu')
```
justinlittman: Should not default to sfm@gwu.edu.  Can either make required or optional.  If optional and not provided, then display is skipped.

justinlittman:
I'd suggest a bit more of a left margin on the footer.  It appears too close to the left edge.

Tanych:
I think it could the browser cache problem. The new css file conflicts with the cache file. if the cached css file has been deleted. It could be like this:
center

Issue #599, opened 2016-12-16
Document export columns for Weibo


kerchner:

lwrubel:
Use Twitter data dictionary as a model for this.

Issue #595, opened 2016-12-15
In documentation, add links to archived social media platforms' data dictionaries


reliztrent:
e.g., http://sfm.readthedocs.io/en/latest/data_dictionary.html could link to:

Twitter API documentation - https://web.archive.org/web/*/https://dev.twitter.com/docs
Tweets - https://web.archive.org/web/*/https://dev.twitter.com/overview/api/tweets (this may have had an alternate URL before January 2015)
Entities - https://dev.twitter.com/docs/platform-objects/entities

lwrubel:
@yonahbg I can give context and explanation for this. See me (or email) for discussion first. Roll in with #586, #587, #599.

lwrubel:
Closing since work is merged into master with data dictionary work.

kerchner:
Closed in #621

Issue #587, opened 2016-12-08
Document export columns for Tumblr


yonahbg:
Create a 'data dictionary' to explain each of the columns in the CSV and Excel exports for Tumblr collections.

Similar to #552 and #586

lwrubel:
See Twitter data dictionary as a model.

Issue #586, opened 2016-12-08
Document export columns for Flickr collections


yonahbg:
Create a 'data dictionary' to explain each of the columns in the CSV and Excel exports for Flickr collections.

Similar to #552

lwrubel:
For data taken field, see #505 for notes about how this field reflects the timezone of the owner.

PR #621, opened 2017-01-05
Ref #586 #587 #595 Tumblr and Flickr Data Dictionaries

yonahbg:
Does not address Issue #599 for Weibo dictionary


yonahbg:
Now includes Weibo

Issue #605, opened 2016-12-20
Clarification of processing docs


justinlittman:
Suggestions from @chrisprom.

On https://sfm.readthedocs.io/en/latest/processing.html#processing:

> This section is very interesting but incomplete.  I assume that in order to use these tools you need to have direct access to the server where SFM is installed and the ability to run shell commands against that instance, at least if you are doing this in the way specified under https://sfm.readthedocs.io/en/latest/processing.html#processing-in-container.   In any case, more detail would help power users like me know more about how to put these processing tools into action.  (Note I have not tried any of this yet since I am still getting enough data on my laptop to make it worthwhile.)

On https://sfm.readthedocs.io/en/latest/processing.html#processing-locally:

> Not sure I am understanding this section.  It seems to be saying I could do processing from my local machine, if I have a linux shell and ability to access to warcs on the SFM server. But not sure I follow this since a bit too technical to me (as someone who has moderate linux and networking knowledge.)

justinlittman:
Also, it was earlier suggested by @lwrubel that Approaches section should come before the Tools section.

Further, should consider getting rid of the processing locally section.

PR #628, opened 2017-01-09
refs #605. Updates and clarifies processing docs.

justinlittman:


yonahbg:
This looks good. I also get the impression very early what skills are required--even the title, adding that it's for command line processing, helps to illustrate the skills needed.

I also was able to understand the introduction, so it was a helpful beginning. I couldn't understand much past then, but it seemed okay.

Issue #574, opened 2016-12-06
Warn admins on long queue lengths


justinlittman:
![screen shot 2016-12-06 at 7 14 22 am](https://cloud.githubusercontent.com/assets/588335/20925239/bd4a36f4-bb83-11e6-89af-369c9b5e3105.png)

lwrubel:
This severely impacts the Twitter rest harvester for all SFM users. Discuss design with @justinlittman before implementation.

Tanych:
@justinlittman 
what I think with this issue is that we can add an `interval` job (can be 2 hour or 4 hour) to check the queue message, if any queue of the request is larger than the setting threshold( might be 100) , we will send a message to the admin.

justinlittman:
So rather than implement multiple checks (e.g., space usage, queue lengths) each of which are scheduled separately, I'd prefer to see a single health check has one schedule, but performs multiple checks.

Tanych:
That's true, one schedule job checks all queue info, if any of them has reached the limit, warning message will sent.

lwrubel:
@justinlittman  ready for testing

PR #633, opened 2017-01-12
Ref #574 adding warn message for queue length

Tanych:
Ref https://github.com/gwu-libraries/sfm-ui/issues/574
adding warning interval job to notification the sfm-admin check the harvester queue
Ref https://github.com/gwu-libraries/sfm-ui/issues/641
adding monitoring sfm-ui

### DIFF
--- a/sfm/sfm/settings/common.py
+++ b/sfm/sfm/settings/common.py
@@ -218,6 +218,28 @@ DATA_THRESHOLD = env.get('DATA_VOLUME_THRESHOLD', '10GB')
 # sfm processing space threshold to send notification email,only ends with MB,GB,TB. eg. 500MB,10GB,1TB
 PROCESSING_THRESHOLD = env.get('PROCESSING_VOLUME_THRESHOLD', '10GB')
 
+# Whether to scan the amount of free space on /sfm-data and /sfm-processing
+PERFORM_MONITOR_QUEUE = env.get('SFM_PERFORM_MONITOR_QUEUE', 'True') == 'True'
+# frequency to check the queue message length,default is 2 hour
+MONITOR_QUEUE_HOUR = env.get('SFM_MONITOR_QUEUE_HOUR', '2')
+# queue threshold for each harvester to send warning message
+QUEUE_LENGTH_THRESHOLD = {
+    'Web Harvester': env.get('WEB_QUEUE_LENGTH_THRESHOLD', '25')
+    # 'Twitter Rest Harvester': '33',
+    # 'Flickr Harvester': '20',
+    # 'Tumblr Harvester': '30',
+    # 'Twitter Harvester': '20',
+    # 'Weibo Harvester': '30',
+    # 'Weibo Exporter': '40',
+    # 'Twitter Stream Exporter': '10',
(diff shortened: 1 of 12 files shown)


sfm/sfm/settings/common.py
```diff
@@ -209,6 +209,13 @@
 # sfm processing space threshold to send notification email,only ends with MB,GB,TB. eg. 500MB,10GB,1TB
 PROCESSING_THRESHOLD = env.get('PROCESSING_VOLUME_THRESHOLD', '10GB')
 
+# Whether to scan the amount of free space on /sfm-data and /sfm-processing
+PERFORM_MONITOR_QUEUE = env.get('SFM_PERFORM_MONITOR_QUEUE', 'True') == 'True'
+# frequency to check the queue message length,default is 2 hour
+MONITOR_QUEUE_HOUR = env.get('SFM_MONITOR_QUEUE_HOUR', '2')
+# queue threshold to send warning message
+QUEUE_LENGTH_THRESHOLD = env.get('QUEUE_LENGTH_THRESHOLD', '100')
```
justinlittman: Set default to 10.

sfm/ui/notifications.py
```diff
@@ -171,6 +172,48 @@ def _create_space_email(email_address, msg_cache):
...
+                log.error("Error sending email: %s", ex)
+
+
+def _create_queue_warn_email(email_address, msg_cache):
+    text_template = get_template('email/queue_length_email.txt')
+    html_template = get_template('email/queue_length_email.html')
+    msg_cache["url"] = _create_url(reverse('home'))
+    msg_cache["monitor_url"] = _create_url(reverse('monitor'))
+    d = Context(msg_cache)
+    msg = EmailMultiAlternatives("[WARNING] Long queue message on SFM server",
```
justinlittman: "Long message queue" not "Long queue message"

sfm/ui/templates/email/queue_length_email.html
```diff
@@ -0,0 +1,17 @@
+{% load humanize %}
+<html>
+<body>
+<p>The following message queues on your <a href="{{ url }}">Social Feed Manager</a> server reach the warning level with <strong>{{ queue_threshold }}</strong> requests. </p>
```
justinlittman: "have reached"

justinlittman:
Upon further reflection, it will be necessary to be able to override on a per harvester/exporter basis.  This is particularly necessary for the web harvester, which should be expected to have a sizable queue length.

I'd recommend this be provided as a map in settings.  By default, the only value in that map can be for the web harvester, which should have it's own env setting.

justinlittman:
One last request:  When there is a queue warning, can you put it on the home page for admins (similar to space usage).

Tanych:
what I understand is that the warning message will only show when it has queue reach the warning level, otherwise it won;t show: it would look like:
new_admin_page

Should more simple?

Tanych:
@justinlittman If it's too complex, I will simplify to show queue name.

Tanych:
@justinlittman you can test the new view of the admin page.

Issue #647, opened 2017-01-18
Docs for resending warccreated messages says "container id" rather than "hostname"


runderwood:
In the ELK docs, the instructions for resending warccreated messages says container id. It appears to me it's actually supposed to be the configured hostname (default: myproject_elk).

justinlittman:
Verified and fixed.  Thanks @runderwood!

Issue #648, opened 2017-01-19
Quoted term in Twitter search seed raises uncaught exception


runderwood:
Entering a term like `"bird law"` raises an uncaught exception in ui/models.py, line 505.

I've fixed it locally by changing line 507 to:

```except (ValueError, AttributeError):```

I can fork and do a pull request if you want. But it's a trivial change. Let me know.

justinlittman:
Our grant would smile kindly upon a pull request.  Thanks!

justinlittman:
Fixed by https://github.com/gwu-libraries/sfm-ui/pull/651

PR #651, opened 2017-01-23
refs. #647. improved exception handling for quoted seeds

runderwood:


runderwood:
this actually refs. #648, which is one more than the ticket number above. sorry!

PR #655, opened 2017-01-27
Fixes #573 - Avoids duplicate credentials

### LINKED ISSUE #573: Credential conflicts
Adding the same twitter credential from different users results in an error
screenshot from 2016-12-05 13 48 37

adityadharne:


justinlittman:
For the sake of not having to repeat code, it would be cleaner to have a method like the following on BaseCredentialForm:

```
    def clean(self):
        cleaned_data = super(CredentialFlickrForm, self).clean()
        token = json.dumps(self.to_token(clean_data))
        if Credential.objects.filter(token=token).exists():
            raise ValidationError('This is a duplicate of an existing credential')
        return cleaned_data
```

Each platform specific CredentialForm would then need to implement to_token(), like:

```
       def to_token(cleaned_data):
             return {
            "key": cleaned_data["key"].strip(),
            "secret": cleaned_data["secret"].strip(),
        }
```

The save() method could also be refactored to use to_token().

PR #697, opened 2017-02-07
Added post announcing 1.5.

justinlittman:


lwrubel:
Changes for your consideration, @justinlittman.

Issue #626, opened 2017-01-09
Admin interface needs to handle large lists


justinlittman:
The Admin interface currently has a number of problems:
1. Some of the form elements in the Admin interface have huge lists.  This makes loading the pages extremely slow.  For example, try a harvest object in production.
2. The Admin interface is configured to search some non-searchable fields.  For example, searching on a seed results in a 500.

This ticket is to clean-up the Admin interface.  This includes:
1.  Making sensible choice about what is displayed in the list, is available is a filter, and is searchable.
2.  Marking fields that will have huge lists as read-only.

PR #699, opened 2017-02-10
refs #626. Cleans up admin interface.

justinlittman:


Tanych:
@justinlittman 
In general, it's good to go for this ticket. 
* [x] all the fields display correctly
* [x] search fields is searchable and non-search is invalid

Performance test for seed pages, other models have the similar situation. 

### Test case for seed
* loading 500+ seeds to one collection
* checking the response

It returns the expected results.

Issue #625, opened 2017-01-09
Skipping not handled properly for Twitter stream collection


justinlittman:
If a Twitter sample or filter stream has a last harvest with the stop requested status (either because it is stuck or because it was just turned off), turning on the collection will result in the next harvest being skipped.  The only way to void the earlier harvest is to use the admin interface.

Twitter stream collections should be exempted from skipping.

justinlittman:
Steps to reproduce:
1.  Create a Twitter filter stream collection and turn it on.  Allow it to perform some harvesting.
2.  Turn it off.
3.  Turn it back on immediately.

Expected result:
User cannot turn back on until previous harvest is completed.

Actual result:
User can turn it on and new harvest is skipped.

PR #701, opened 2017-02-10
refs #625. Fixes toggling streams on and off.

justinlittman:


Tanych:
@justinlittman 
It might be a little complicated for the second test case, but if it occurs, it would cause trouble.

## Test case for status `STOP_REQUESTED`
* building a twitter filter collection
* turn on the collection 
* turn off the collection in 2 minutes
* turn on button disabled and refresh the page, the collection completed success
* turn on button enabled and turn it on, the collection runs successfully.

It's good for this case

## Test case for status `PAUSED`(container exits unexpected)

**To follow up the following process, you need to cool down with a fresh mind.**

* Building a twitter filter collection with `trump`
* Turn on the collection and running collection for 1-2 minutes
* Kill the container and **assume the container can't restart automatically due to unexpected reason**
The status is still marked as `running`.
kill_container_status
I am not sure whether the status should be marked as `paused`.

* Right now, if the user turns off the collection and the status marked as `stop requested`.
* Restart the container (assume you have fixed the issue and the harvester container can restart), the collection status is turned off but the harvester status is `running`.
turn_on
status_running
**The status don't match for each other.**

* Turn on the collection, but requests have been skipped
status_skipped
* Try to void the harvester which is pending running and turn the collection off and then turn it on.
The next harvests would be pending on `request`
pending_requests

No matter what you do, even restart the containers, the twitter harvester can't run normally. I am not sure whether it should consider as a problem. I think the key problem is that when the harvester containers exit unexpected, the sfm-ui should mark the status as `paused` and disable the turned off button like the `streaming stop`.

> If you can't follow above steps, I can reproduce the problem on your side.

Issue #712, opened 2017-02-23
Seed warning messages


justinlittman:
As suggested by @kerchner for Program on Extremism, it would be really helpful if seed warning messages were displayed next to the seeds to which they apply on the collection detail page.

Related to #706.

justinlittman:
PRs for this ticket:
* https://github.com/gwu-libraries/sfm-ui/pull/713
* https://github.com/gwu-libraries/sfm-utils/pull/31
* https://github.com/gwu-libraries/sfm-twitter-harvester/pull/35
* https://github.com/gwu-libraries/sfm-flickr-harvester/pull/17
* https://github.com/gwu-libraries/sfm-tumblr-harvester/pull/12

Issue #706, opened 2017-02-16
Notify user of dupe seeds


justinlittman:
There are cases in which updates from harvests can result in duplicate seeds for a collection (meaning that more than one seed has the same token or uid).

The user should be notified on the collection details page when there are duplicate seeds.  The warning should to at the seed level, not the overall collection level (to make figuring out which seeds to fix easier).

Related to #640 and ticket for putting harvest warnings inline with seeds.

kerchner:
Fixed in #713.

PR #713, opened 2017-02-23
refs #712, #706. Adds seed message for dupes and harvester warnings.

justinlittman:


kerchner:
I ran a harvest that returned the following warnings:
```
    User id not found for user _0Die_Lowin_9 because account is not found or suspended
    Screen name not found for user id 777455042106363904 because account is not found or suspended
    Screen name not found for user id 771102208389705728 because account is not found or suspended
    Screen name not found for user id 771436744260456448 because account is not found or suspended
    Screen name not found for user id 781235309547757568 because account is not found or suspended
    Screen name not found for user id 777455042106363904 because account is not found or suspended
    Screen name not found for user id 759476376097488896 because account is not found or suspended
```
and those messages were not shown in the seed list, whereas I would have expected them to.

However, each seeds which was a duplicate did display the `Duplicate seeds exist with this token` message.

justinlittman:
@kerchner I suspect that you weren't running the correct branches from sfm-utils and sfm-twitter-harvester.  See https://github.com/gwu-libraries/sfm-ui/issues/712#issuecomment-282063985.

Also, I've rebased this branch so a normal pull won't work correctly.

screen shot 2017-03-01 at 7 35 58 am

kerchner:
@justinlittman Indeed this was because I didn't have the other repos on the right branches - thanks for pointing me to the comment in the issue (though I would tend to look in the PR first for setup instructions needed to test the PR).  This works well, although when refreshing the collection details page, if you time it just right, there's a brief period where the Messages column is blanked out.  If that was its own issue, it would be super low priority, hence merging.

Issue #640, opened 2017-01-14
Seed duplicate key problem


justinlittman:
From EOT instance:

```
ui_1                      |   File "/usr/local/lib/python2.7/site-packages/django/db/utils.py", line 98, in __exit__
ui_1                      |     six.reraise(dj_exc_type, dj_exc_value, traceback)
ui_1                      |   File "/usr/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
ui_1                      |     return self.cursor.execute(sql, params)
ui_1                      | django.db.utils.IntegrityError: duplicate key value violates unique constraint "ui_seed_collection_id_5a9ae0d72334738_uniq"
ui_1                      | DETAIL:  Key (collection_id, uid, token)=(1, 3302915626, FortLarnedNPS) already exists.
```

![screen shot 2017-01-14 at 3 31 27 pm](https://cloud.githubusercontent.com/assets/588335/21957867/9ee6968a-da6e-11e6-960f-4d71fa8fc12c.png)

justinlittman:
Steps to reproduce:
1.  Create a twitter user timeline credential.
2.  Add the seeds "FortLarnedNPS" and "fortlarnednps".
3. Turn on the collection.
4. After the first harvest is completed, turn off and back on so that a second harvest is performed.
5. Observe the logs of the ui consumer.

Expected result:
At step 2, it should not have been possible to enter the same twitter handle with different capitalization.

Actual result:
It was possible to enter the same twitter handle with different capitalization.  And it caused a db error when attempting to set the two seeds to have the same uids and tokens.

I suspect that there is a similar problem if a collection was created with 2 seeds, one with uid "3302915626" and one with gid "FortLarnedNPS".

Unfortunately, there is no way to make Django's unique together constraint case insensitive.  I'd suggest that we need to undo the unique_together constraint and add case-insensitive validation during data entry (both when adding individual and bulk seeds).  This will still allow some duplicates (e.g., in the second example given above), but would address the general problem.

lwrubel:
Seed approach taken for credential conflicts.

kerchner:
Any duplicate checking is limited to check against Active seeds only.

justinlittman:
Here are the requirements for this ticket:

[...]

Issue #714, opened 2017-02-24
seed duplicate problem in other collection type


Tanych:
The seed duplicate and validation problem should do for the following collection type:
* Flickr user : checking the scree name and NSID, if possible, specify the NSID rules.
* Tumblr post: screen name should not  be number and empty.
* Twitter filter: track, follow, location as one seed.
* Twitter search: keyword should not be empty.

if possible, for 
* weibo search

justinlittman:
Reconsider this ticket after suggestions from review of #640 are implemented.  I suspect most of this additional validation will prove unnecessary.

Tanych:
@justinlittman  this tickets is not necessary, I will update an all in one check duplicate seed in #640.
closed!

PR #715, opened 2017-02-24
Ref #640 #727 seed duplicate for twitter user timeline

Tanych:
Adding validation rule for twitter user timeline.
* checking duplicate when create and update
* checking null when create and update
* allow udpating status 

Ref #640, For detail test cases, please check the issue page.
Ref #727, seed value error

### DIFF
--- a/sfm/ui/forms.py
+++ b/sfm/ui/forms.py
@@ -476,6 +476,10 @@ class BaseSeedForm(forms.ModelForm):
 
     def __init__(self, *args, **kwargs):
         self.collection = kwargs.pop("collection", None)
+        # for createView and updateView
+        self.view_type = kwargs.pop("view_type", None)
+        # for updateView check the updates for the original token and  uid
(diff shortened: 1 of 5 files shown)


kerchner:
Also note related issue #714

Tanych:
@justinlittman  it's ready for test!

justinlittman:
In this case (1) "One of the following fields is required :Screen name User id." shouldn't be displayed and (2) the wording should be correct ("One of the following fields is required: Screen name or User id.")

screen shot 2017-03-06 at 3 05 28 pm

justinlittman:
After that fix and a rebasing, it is good to go.

Tanych:
@justinlittman fixed, if it's good to go, please merge and close

kerchner:
@justinlittman @Tanych Before merging, note the case I added to #640 - may require additional work (or not, see what you think).

Issue #630, opened 2017-01-10
Add configuration to hide/unhide Weibo search collection type


kerchner:
Since the Weibo search API requires credentials with advanced access, presenting Weibo search as a collection type might be confusing to most users, as most users with Weibo credentials are unlikely to have advanced access.

Add a configuration in the environment variables to (in the UI) show/hide Weibo collection types that require advanced API access (right now this is only the Weibo Search collection type - see https://github.com/gwu-libraries/sfm-ui/pull/624). By default they should be hidden.

kerchner:
When adding this variable to `example.env` (in sfm-docker), provide sufficient comments above the variable to explain it, with links to Weibo (and/or SFM) documentation about Weibo advanced API access.

lwrubel:
Per discussion with @kerchner, this may move into the next milestone if more time needed for review.

kerchner:
Fixed in https://github.com/gwu-libraries/sfm-docker/issues/17 and in #624

PR #723, opened 2017-02-28
Ref #724 adding doc for weibo search

Tanych:
adding Weibo search collection for readthedocs 
Ref #630 Ref #724

### DIFF
--- a/docs/collections.rst
+++ b/docs/collections.rst
@@ -17,6 +17,7 @@ Collection types
     tweets in real time.
   * `Flickr user`_: Collects posts and photos from specific Flickr accounts
   * `Weibo timeline`_: Collects posts from the user and the user's friends
+  * `Weibo search`_: Collects recent weibo posts by a user-provided search query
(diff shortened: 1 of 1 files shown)


kerchner:
@Tanych I have a couple of questions:
- What is the '#' on Weibo posts that you're referring to?
- It sounds like when `incremental` is selected, the Weibo Search does another call to reduce the number of duplicates.  I'm not sure I understand how the additional API call reduces duplicates  (versus what it does when `incremental` is not selected).

Tanych:
@kerchner 
1. the topic or keyword is between the `#` in Weibo posts, like #keywords#, #你好#. 
2. if `incremental` is selected, it will record the max weibo id has been harvested, if the weibo id is less than the recorded id, it won't collect. The duplicate action is to request an addition API request to check whether it has new Weibo posts because the search API has no `since_Id` and `max_id`. However, the Warc file will record the additional API requests, so it could be a little duplicate.

kerchner:
@Tanych I pushed some edits.  Is this close to what you meant to say, regarding the incremental weibo search harvest?

Issue #725, opened 2017-02-28
Weibo search collection description


Tanych:
Due to the UX update, the weibo search need to add detail description for the weibo search collection.
* Add Weibo search: Recent Weibo posts matching a query

kerchner:
Fixed in #726

PR #726, opened 2017-02-28
Ref #725 adding weibo description

Tanych:
Ref #725, adding weibo search collection description and remove the hard code for the UX for collection description.

### DIFF
--- a/sfm/ui/config.py
+++ b/sfm/ui/config.py
@@ -46,6 +46,7 @@ class UIConfig(AppConfig):
         if settings.WEIBO_SEARCH_OPTION:
             log.debug("Adding weibo search collection")
             Collection.HARVEST_CHOICES.append(('weibo_search', 'Weibo search'))
+            Collection.HARVEST_DESCRIPTION.append(('weibo_search', 'Recent Weibo posts matching a query'))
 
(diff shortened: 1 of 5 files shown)


Tanych:
@justinlittman  can you take 2 minutes to review this PR? If it's OK, merge it and close the ticket.

Tanych:
@adityadharne Well, I think It makes more general and flexible to add harvester type. Weibo harvester is a special case that it only turns on when people want to. Also, it won't alter model and do migrations, it's just a class member. As the general process, all the process should do on the view level. The templates are hard to control the process and back-end logical.

adityadharne:
@Tanych No, I meant this can be done in a later ticket- There are alterations to that ticket which are not in master yet. It will cause merge conflicts.

Tanych:
@adityadharne OK, you can go ahead, and  I will fix the conflicts if necessary.

kerchner:
Dependent on merging #731 .  May requires rebase.

kerchner:
@Tanych Requires rebase (probably due to my merging #713)

Tanych:
@kerchner  It could be #731. The UX change. Now, it fixed.

Issue #708, opened 2017-02-17
Documentation on publishing dataset to Dataverse


justinlittman:
Provide detailed instructions on the steps for publishing a dataset to Dataverse.  The Women's March collection can be used as an example.

lwrubel:
Also a short blog post on the context and use for the new documentation.

justinlittman:
Here's the PR for the blog post: https://github.com/gwu-libraries/sfm-ui/pull/750

PR #744, opened 2017-03-13
Ref #708 dataverse docs

soominpark:
Please feel free to edit or comment on this document. 

I applied all Laura's suggestions (replacing the production SFM server name to a fake one, adding a step to disconnect from the container before and after step 5, replacing the Dataverse link in step 1 that was not working to a new one, removing port 9999, and correcting some typos) except for making the Dataverse link open in a separate window since I didn't figure it out yet.


soominpark:
@lwrubel Thank you for your review! It's been a big help. I will update the images whenever the new UI is available.

Issue #749, opened 2017-03-14
Show all seeds when clicking "see all"


lwrubel:
"See all n seeds" link at bottom of seeds list on collection detail page should cause accordion to show all seeds. Currently it is only showing the next 10 seeds and shows ten more with each click.

yonahbg:
@adityadharne Thanks, I didn't see this

adityadharne:
PR: https://github.com/gwu-libraries/sfm-ui/pull/766

Issue #738, opened 2017-03-09
Option to hide web harvests in harvests list 


lwrubel:
Add an option to filter harvest list to type, filter status at a minimum.

Use something like GitHub's limit options.

adityadharne:
PR: https://github.com/gwu-libraries/sfm-ui/pull/766

justinlittman:
This doesn't work quite right.

First and most importantly, it only limits against the harvests that are listed in the table, not the list of all harvests.  If I select, "Completed with errors" and none of the last 5 harvests completed with errors then nothing will be displayed. However, I may have harvests that completed with errors -- they're just not in the last 5.

Second, search doesn't make sense for limiting by harvest type. For any give collection, there will only be 2 harvest types: the collection's type and web harvest.

Third, once you get to the rest of the harvests (click View all XX harvests), there is no ability to limit.

Because the first is a defect, I'm removing from this release.

lwrubel:
This will be simpler to implement in the harvests ListView. Set up a filter there (same dropdown as status?). Change link at bottom of Harvests table to "View all n harvests or filter further".

justinlittman:
At this point, we're not doing many web harvests so I'm going to remove this ticket from milestone.

justinlittman:
Closing since we're not doing many web harvests.

Issue #694, opened 2017-02-06
Reconsider Active label for seeds


lwrubel:
Applies to seed lists in collection detail page and seed detail page. "Include in next harvest?" or something else.

lwrubel:
Add option to filter to show "deleted" seeds. Change "Active" option in seed detail page to "Delete".

adityadharne:
PR: https://github.com/gwu-libraries/sfm-ui/pull/766

Issue #685, opened 2017-02-06
Collection detail page: Define “Seeds” in context of collection type


lwrubel:
Provide a collection-type specific definition of seed in the panel heading. 
* Seeds (Twitter accounts)
* Seeds (Filter criteria)
* Seeds (Search query)
* Seeds (Tumblr blogs)
* Seeds (Flickr users)
* Seeds (Weibo users)

Place a question mark bubble graphic (check font awesome) next to Token and UID and provide a pop-up/modal explanation of each of those, with an example.

Instead of “Token” and UID, use label appropriate for the type of collection (e.g. username, user ID, blog name).

Hide the Token column in the case of Tumblr, since it's always empty.

lwrubel:
Place a question mark bubble graphic (check font awesome) next to Token and UID and provide a pop-up/modal explanation of each of those, with an example. 

Instead of “Token” and UID, use label appropriate for the type of collection (e.g. username, user ID, blog name).

lwrubel:
Hide the Token column in the case of Tumblr, since it's always empty.

adityadharne:
PR: https://github.com/gwu-libraries/sfm-ui/pull/766

PR #766, opened 2017-03-30
Adds filters and search to collection detail page

adityadharne:
Fixes: 
- https://github.com/gwu-libraries/sfm-ui/issues/749
- https://github.com/gwu-libraries/sfm-ui/issues/738
- https://github.com/gwu-libraries/sfm-ui/issues/694
- https://github.com/gwu-libraries/sfm-ui/issues/685
- https://github.com/gwu-libraries/sfm-ui/issues/526

### DIFF
--- a/sfm/ui/forms.py
+++ b/sfm/ui/forms.py
@@ -446,7 +446,7 @@ class CollectionTumblrBlogPostsForm(BaseCollectionForm):
 class BaseSeedForm(forms.ModelForm):
     class Meta:
         model = Seed
-        fields = ['collection', 'is_active',
+        fields = ['collection',
                   'history_note']
         exclude = []
         widgets = {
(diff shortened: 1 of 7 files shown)


lwrubel:
@adityadharne a few preliminary comments as I'm working through the PR:
* Rather than putting the lengthy jQuery code in the collection_detail template, use a static js file. (https://docs.djangoproject.com/en/1.8/intro/tutorial06/). This will make the template cleaner.
* Can you put the harvest type dropdown options into the context passed from the view? There are hardcoded types in the template currently, and the types might change in the future. 
* A general usability principle is to avoid placeholder text in search boxes unless absolutely needed. Label should suffice for the seeds and harvest types search boxes: "Search seeds:" and "Search harvest types:".
* The panel header for the Seeds panel has the number of seeds on the same line as the heading. Would you make the harvests panel similar? This will make the panel header less large.

PR #771, opened 2017-04-06
Added releasing 1.7 blog post.

justinlittman:


lwrubel:
@justinlittman Take a look at the edits and make sure you're comfortable with them. Is it tough to tweak the image of the arrow pointing to "Add note and turn off" so that the arrow points to the adding note option?

PR #779, opened 2017-04-11
Fixes #773 - clarifies seed counts

### LINKED ISSUE #773: Clarify seed counts
1. Where ever seed counts appear, it should be clarified if active, deleted or total.
2. If only displaying 1 number, I'd suggest that active makes the most sense.
3. On seed list, would be helpful to have breakdown of active/deleted.

adityadharne:


lwrubel:
@adityadharne badges look good. If the count is 0, however, no badge should show. Bootstrap automatically hides the badge if it's empty, so you could put logic in the template to show nothing in that case.

Issue #769, opened 2017-04-04
Put an "add note" button on collection set detail and seed detail 


lwrubel:
See button at top of Change log on collection detail page as an example.

PR #786, opened 2017-04-20
refs #769. Adds note button to collection set detail and seed detail …

justinlittman:
…pages.


rithvikmundra:
@justinlittman I tested the above changes and it works fine.

PR #791, opened 2017-04-24
Fixes #661 - Broken Tables and Change Logs fixed

### LINKED ISSUE #661: Long track filter busts tables
Keep tables from breaking when there's a long string for the twitter filter seed field. 

screen shot 2017-01-31 at 4 47 20 pm

rithvikmundra:


lwrubel:
As discussed, apply this style to seed_detail.html as well. Since this is in a bunch of places at this point, create a style in main.css and add that to the templates instead of inline styles.

lwrubel:
@rithvikmundra I think there is also something going on with the bootstrap rows, with a row nested within another, and thus making you put fixed widths on the table columns. If you can check the rows, you may be able to get rid of that part of the html.

rithvikmundra:
@lwrubel I made the necessary modification. Please verify the same.

rithvikmundra:
Closing this pull request so that I can delete this branch and can start from scratch.

PR #792, opened 2017-04-25
Fixes #716 - Changes to user harvest emails

### LINKED ISSUE #716: Daily email info for harvests that are less frequent than daily
Daily emails for harvests that are weekly (or less frequent than daily) show 0 for harvests "Yesterday" and "Previous day". This makes it look like there were no tweets collected or perhaps there was a problem, when in fact no harvest was scheduled for that day.  

```
Item	 Yesterday	Previous day	Last 7 days	
tweets	0	0	933	1,609	
```

Change 0 to N/A (or something similar) when there was no harvest scheduled that day.

adityadharne:


justinlittman:
This doesn't correctly handle the case when there was a harvest on the previous day, but it returned no results.  This code would display it as N/A instead of 0.

lwrubel:
This is being closed without merging because handled in #799.

Issue #794, opened 2017-04-25
Handle rogue stream harvests


justinlittman:
I've seen cases in production where stream harvests are continuing, even though a stop has been requested.

To deal with this, if the ui consumer receives a status message for a harvest that is stop requested or void, it should issue a harvest stop message. Redundant harvest stop messages aren't harmful.

justinlittman:
This ticket depends on:
* https://github.com/gwu-libraries/sfm-ui/pull/795
* https://github.com/gwu-libraries/sfm-utils/pull/34

Steps to test:
1. Start a filter harvest.
2. Using Admin, change the status of the harvest to STOP REQUESTED.
3. Wait 30 minutes.
4. Verify that the status of the harvest is SUCCESS.

PR #795, opened 2017-04-26
refs #794. Adds STOPPING harvest status. Checks state of RUNNING harv…

justinlittman:
…ests to look for rogue streaming harvests.

### DIFF
--- a/sfm/message_consumer/sfm_ui_consumer.py
+++ b/sfm/message_consumer/sfm_ui_consumer.py
@@ -54,6 +54,8 @@ class SfmUiConsumer(BaseConsumer):
                       json.dumps(self.message, indent=4))
             return
 
+        current_harvest_status = harvest.status
+
         # And update harvest model object
         harvest.status = self.message["status"]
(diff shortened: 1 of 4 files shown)


Tanych:
@justinlittman 
### Stop Request
* create a twitter filter collection
* running for 1 or 2 minutes
* turn off the collection
* collection marked as `STOP REQUEST` and immediately return successfully with tweet status.
### Void Request
* create a twitter filter collection
* running for 1 or 2 minutes
* turn off the collection
* collection marked as `Void` 
* It takes about 30 minutes to return successfully with tweet status.

Why the `Void` operation take so long time?  I didn't see any process to take so much longer time than `STOP REQUEST`.

justinlittman:
Stop sends an immediate stop message to the harvester; void waits until it gets a status message from the harvester and then sends a stop message.

The void behavior is only a backup for when something goes wrong.

Tanych:
@justinlittman I have retested it, it seems that it's good to go.

PR #797, opened 2017-04-27
refs #786. Adds check on serialization that correct collection.

justinlittman:
I can't think of a good way to test this.

### DIFF
--- a/sfm/ui/serialize.py
+++ b/sfm/ui/serialize.py
@@ -214,7 +214,9 @@ class RecordSerializer:
     def _historical_seeds_iter(collection):
         for seed in collection.seeds.all():
             for historical_seed in seed.history.all():
-                yield historical_seed
+                # For undetermined reasons, sometimes get historical seeds that are not part of collection.
+                if historical_seed.instance.collection == collection:
+                    yield historical_seed
 
     def _serialize_harvests(self, collection, records_path):
         # Harvests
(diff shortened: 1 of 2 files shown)


Tanych:
@justinlittman I will test it.  Also, whether should take consider set the `ignorenonexistent` parameters to `true` in  Django `serializers.deserialize` ?  
https://docs.djangoproject.com/en/1.11/topics/serialization/

justinlittman:
So since it is probably an error if fields are missing on import, I think we should not set ignorenonexistent.

Tanych:
@justinlittman I compared all the JSON files on the records generates by current PR and 1.7.0, only the `historical_seeds.json` has some key diff.
```
Diff in historical_seeds.json
17623a17624,17645
>         "history_type": "+",
>         "uid": "",
>         "date_updated": "2016-10-04T15:39:54.597Z",
>         "is_active": true,
>         "collection": [
>             "8b9627b317ca4b6eb333b844d97001d5"
>         ],
>         "history_note": "",
>         "token": "{\"track\": \"#debate, #debates, #debatenight, #debate2016, #debates2016, vice-presidential debate, vice presidential debate, vp debate\", \"follow\": \"22203756, 172858784\"}",
>         "is_valid": true,
>         "seed_id": "e8de2f8bff114712b6e2033d6e6a214d",
>         "date_added": "2016-10-04T15:39:54.594Z",
>         "history_user": [
>             "exampleuser"
>         ],
>         "id": 2700,
>         "history_date": "2016-10-04T15:39:54.600Z"
>     },
>     "model": "ui.historicalseed"
> },
> {
>     "fields": {
Diff in info.json
2c2
<     "serialization_date": "2017-05-03T11:30:38.664127-04:00",
---
>     "serialization_date": "2017-05-03T11:09:42.484284-04:00",
Diff in seeds.json
Diff in users.json
Diff in warcs.json

```

It's  good to go!

PR #799, opened 2017-04-28
Changed notifications to put N/A when no harvests in time period. Add…

justinlittman:
…ed some time humanization in emails.


adityadharne:
@justinlittman this works but travis is failing.

justinlittman:
Verified locally that unit tests pass.

Issue #800, opened 2017-05-01
Missing humanized numbers


justinlittman:
There are a handful of places where numbers are not humanized. They drive me crazy.

![2017-05-01_1211-2](https://cloud.githubusercontent.com/assets/588335/25585256/ab7c1fee-2e67-11e7-91fa-ced186f848a2.png)
![2017-05-01_1211](https://cloud.githubusercontent.com/assets/588335/25585257/ab80aeba-2e67-11e7-805f-a3675bf3ff34.png)
![2017-05-01_1209](https://cloud.githubusercontent.com/assets/588335/25585258/ab80ccc4-2e67-11e7-9c60-5f24ca95dde8.png)

justinlittman:
PR is https://github.com/gwu-libraries/sfm-ui/pull/801

PR #801, opened 2017-05-01
refs #800. Added commas where missing in numbers.

justinlittman:


Tanych:
@justinlittman 
I rebased from the master, solved the conflicts and add the humanize for the message count. It would be like this:
* Warc
 **Data collected: 13,086 files (21.4 GB)**
* harvester:
havesters_1
* seeds
seeds_1
* change log
change_1

It's good to go!

PR #802, opened 2017-05-01
Fixes #622 - Adds download seed list option for the collection detail…

### LINKED ISSUE #622: Export list of seeds from a collection
It would be helpful to be able to export a list of seeds from a collection. In updating the End of Term collection, for example, I'd like to compare what we have with the US Digital Registry. This would also be helpful in sharing with researchers what we are collecting (e.g. list of news outlets).

adityadharne:
… page

### DIFF
--- a/sfm/ui/templates/ui/collection_detail.html
+++ b/sfm/ui/templates/ui/collection_detail.html
@@ -8,6 +8,7 @@
 {% endblock %}
 
 {% block content_header %}
+
 <div class="row">
   <div class="col-md-12">
       <ol class="breadcrumb">
@@ -192,7 +193,7 @@
 <div class="row">
     <div class="col-md-12">
       <div class="panel panel-default">
-        <div class="panel-heading"><h4>Seeds</h4></div>
+        <div class="panel-heading"><h4>Seeds<a type="button" style="position: relative; top: -6px" class="pull-right btn btn-primary" href="{% url "download" collection.pk %}"><span class="glyphicon glyphicon-download-alt"></span>Download seed list</a></h4></div>
         <div class="input-group"> <span class="input-group-addon">Search</span>
             <input id="filter" type="text" class="form-control">
         </div>
--- a/sfm/ui/views.py
+++ b/sfm/ui/views.py
@@ -30,6 +30,7 @@ from .auth import CollectionSetOrSuperuserOrStaffPermissionMixin, CollectionSetO
 import os
 import logging
 import csv
+import codecs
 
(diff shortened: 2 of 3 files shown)


sfm/ui/views.py
```diff
@@ -196,10 +197,26 @@ def get_context_data(self, **kwargs):
                                                     harvest__historical_collection__id=self.object.id).exists()
         context["item_id"] = self.object.id
         context["model_name"] = "collection"
-
         context["status_choices"] = Harvest.STATUS_CHOICES
         return context
 
+def download(request, pk):
```
justinlittman: Can you make the method name slightly more descriptive, like `download_seed_list`?

sfm/ui/templates/ui/collection_detail.html
```diff
@@ -192,7 +193,7 @@ <h4 class="modal-title" id="addNoteTurnOffModalLabel">Add note & turn off</h4>
 <div class="row">
     <div class="col-md-12">
       <div class="panel panel-default">
-        <div class="panel-heading"><h4>Seeds</h4></div>
+        <div class="panel-heading"><h4>Seeds<a href="{% url "download" collection.pk %}"><span class="pull-right glyphicon glyphicon-download-alt"></span></a></h4></div>
```
justinlittman: Per @lwrubel, please make this a button with a text label like “Download seed list”.

sfm/ui/views.py
```diff
@@ -196,10 +197,26 @@ def get_context_data(self, **kwargs):
...
+    collection = get_object_or_404(Collection, pk=pk)
+    response = HttpResponse(content_type='text/csv')
+    response['Content-Disposition'] = 'attachment; filename="seedlist.csv"'
+    writer = csv.writer(response)
+    writer.writerow([
+        u"Token",
+        u"Uid",
+        u"Link",
+    ])
+    for seed in collection.seeds.all():
```
justinlittman: Should only download active seeds.

sfm/ui/views.py
```diff
@@ -196,10 +197,26 @@ def get_context_data(self, **kwargs):
...
+        writer.writerow([
```
justinlittman: This chokes on a seed like "#美联航强制乘客下机# #北美突发#【美联航与受伤医生达成和解】美联航刚刚与受伤乘客David Dao 达成和解，和解金额仍保密"

Tanych:
@justinlittman @adityadharne another issue need to take consideration is the user level, the users can't directly type the link to download other collections seeds which have no permission to access.

Issue #805, opened 2017-05-01
Weibo search collection_detail page has errors


lwrubel:
@Tanych has identified problems with the UI for Weibo search that prevent it from loading properly.

Tanych:
The Weibo search would like this:
![weibo_search](https://cloud.githubusercontent.com/assets/3956411/25592958/cacfa8c2-2e88-11e7-941b-ba6fcbc12b8e.png)
I will update it with all the stuff related to Weibo and make it fit all the harvester type including all the different names on three columns `link`(some of them might have, some would disappear) `uid` and `token`. It doesn't need to add so much code on HTML templates.

PR #806, opened 2017-05-01
Ref #805 tweak on collection detail for weibo search

Tanych:
Ref https://github.com/gwu-libraries/sfm-ui/issues/805
Tweak collection detail page with multip hard code for collection type!


Tanych:
@justinlittman @lwrubel please consider merging the PR to 1.8 since it can make sure the collection detail page work well for all collection types including `Flickr user` and `Weibo search`.

PR #813, opened 2017-05-04
Fixes #661 - Long Filter Busts Table

### LINKED ISSUE #661: Long track filter busts tables

rithvikmundra:


lwrubel:
@rithvikmundra please apply to seeds table in collection_update.html as well.

lwrubel:
One final tweak: the longseed class is not needed for tables, so it doesn't need a width set. Bootstrap will take care of the width, all that's needed is wordbreak.

Tanych:
We can apply the change to all tables. Make all the table as Fixed table layout. Add css with table:
```css
table {
    table-layout: fixed;
    word-wrap: break-word;
}
```
Same idea with the token filed.

PR #816, opened 2017-05-08
Documentation for AWS template: Host Docker Volume saves to .env

niccdias:
Edits cloud-config template so data & processing volumes are shared with the Docker Host (rather than echoing the instruction to do so into oblivion).

This makes sure the volume instruction (already present) is like all the other instructions being echoed to `.env`


justinlittman:
Thanks @niccdias!

Issue #819, opened 2017-05-11
Consider dormant collections


justinlittman:
As predicted by our use cases and born out by practical experience, we now have a significant number of collections that are dormant, i.e., collecting is completed but the collection is to be retained. An example of this is our election collections.

These dormant collections clutter up parts of the UI and harvest stats emails. It would be optimal to be able to treat dormant collection differently.

Possible approaches:
* A dormant flag for a collection or collection set.
* Automatically treating collection/collection sets as dormant based on inactivity.

lwrubel:
Check with @reliztrent @save4use Vakil about terminology to use. Initial approach is to set manual flag at collection level. Collection set with no active collections is automatically dormant.

justinlittman:
@lwrubel This will need a UI review to make sure implemented appropriately.

PR #826, opened 2017-05-24
refs #819. Changes collection is_active to collection is_on.

justinlittman:
Adds is_active flag to collection. Adds support in UI and harvest status emails.

### DIFF
--- a/sfm/api/serializers.py
+++ b/sfm/api/serializers.py
@@ -11,4 +11,4 @@ class WarcSerializer(HyperlinkedModelSerializer):
 class CollectionSerializer(HyperlinkedModelSerializer):
     class Meta:
         model = Collection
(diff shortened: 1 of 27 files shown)


justinlittman:
This changes the existing is_active flag on Collection to is_on and creates a new is_active flag on Collection. is_active set to false indicates that the collection is dormant.

The following should be tested:
* On the collection detail page, a collection can only be deactivated if it is turned off. When it is deactivated it cannot be turned on or edited, but still can be exported.
* The status is listed as inactive on the collection set detail page.
* Inactive collection sets (i.e., having no active collections) are listed as inactive on the collection set list page.
* Inactive collection sets are omitted from the home page.
* Inactive collection sets and collections are omitted from the harvest status emails.

adityadharne:
This is good to go.

PR #828, opened 2017-06-01
Post annotated bibliography

lwrubel:


lwrubel:
Legal bibliography and related blog post.

Issue #829, opened 2017-06-01
Seeds can be deleted when collection is on


Tanych:
@justinlittman @lwrubel 
Just to make sure whether it is intended to be. 

Turn on a collection with seeds, go to any seeds page in the collection.
The delete operation is allowed when the collection is on.

justinlittman:
@Tanych You are correct. It should not be possible to add/delete/change seeds while collection is turned on.

PR #830, opened 2017-06-02
Ref #829 tweaks on the UI updates

Tanych:
These PR focuses on coding flexible and readable. Some tweaks need to be tested on Chrome mobile and desktop mode with inspecting.

Ref https://github.com/gwu-libraries/sfm-ui/issues/661
Ref https://github.com/gwu-libraries/sfm-ui/issues/829

* tweak on the delete operation when collection is on
* tweak on long seed UI flexible on mobile mode
* tweak on table with flexible on mobile and desktop

### DIFF
--- a/sfm/ui/static/ui/css/main.css
+++ b/sfm/ui/static/ui/css/main.css
@@ -116,16 +116,11 @@ body {
     top: 35%;
 }
 
-
-.longname{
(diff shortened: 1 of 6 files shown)


adityadharne:
This is good to go.

Issue #812, opened 2017-05-04
Send email to admins when error in sfm-ui and sfm-ui-messageconsumer


justinlittman:
Make sure to look into existing functionality in django.

justinlittman:
See https://docs.djangoproject.com/en/1.8/howto/error-reporting/

PR #834, opened 2017-06-12
refs #812. Sends email to admins when error in ui or uiconsumer.

justinlittman:


justinlittman:
Steps for testing in UI:
1. Add an error in a view. (It can just be some random text.)
2. As a user, invoke that review.

Expected result:
1. User receives 500 email.
2. Admin receives email with error.

Steps for testing in UI consumer:
1. Add an error to a message handler.
2. Send a message that invokes that handler.

Expected result:
1. UI consumer terminates.
2. Admin receives email with error.

Issue #835, opened 2017-06-13
Duplicate seed allowed in bulk add


adityadharne:
Duplicate seeds with different capitalizations is allowed and the harvest completes. 
- Example: `adityadharne` and `AdityaDharne` both should not be added.
- Screenshot of the result:

![screen shot 2017-06-13 at 11 55 03 am](https://user-images.githubusercontent.com/15873333/27092194-8fe45978-5030-11e7-888a-cf1125e26b6c.png)

- Previously occurred in https://github.com/gwu-libraries/sfm-ui/issues/640

lwrubel:
Merged with #836.

PR #836, opened 2017-06-13
Ref #835 Fix the insensitive case check duplicate for bulk

Tanych:
Ref #835 

## Test case 
* test, Test, TEST, TeST
* TAN12, tan12, Tan12
* TanYch, tanYCH, TANYCH


adityadharne:
Good to go.

Issue #846, opened 2017-06-21
ELK supporting Weibo topic, user mentions and urls load


Tanych:
currently, no field in Weibo statues can get those data. To solve this problem, using regex to matching these and load them into the elastic search.

lwrubel:
This work supports the CEAL paper and grant work being completed.

Tanych:
@justinlittman PR related
https://github.com/gwu-libraries/sfm-ui/issues/850
https://github.com/gwu-libraries/sfm-weibo-harvester/issues/30
https://github.com/gwu-libraries/sfm-utils/issues/36
https://github.com/gwu-libraries/sfm-docker/issues/20
https://github.com/gwu-libraries/sfm-elk/issues/12

### Test1
weibo data visualization 
* set dev option for logstash
* pull elk image and setting up for elk-instance
* build a weibo timeline collection
* check the dashboard 

### Test2
Load warc weibo data to ELK
* get a weibo warc file 
* run a bash shell into the `sfm logstash` container
* run the following command
```bash
weibo_warc_iter.py --elkwarc=true <your_warc_file> |/usr/share/logstash/bin/logstash -f stdin.conf
```

### Test3
Load JSON weibo data to ELK
* get a weibo full json file exported from the sfm-ui
* run a bash shell into the `sfm logstash` container
* run the following command
```bash
weibo_warc_iter.py --elkjson=true <your_json_file> |/usr/share/logstash/bin/logstash -f stdin.conf
```

PR #850, opened 2017-06-22
Ref #846 update kibana doc

Tanych:
Ref #846 
Add set up doc to make user known the situation!


justinlittman:
Make sure to see other PRs listed in ticket.

Issue #853, opened 2017-06-26
Improved UX for tables in UI


lwrubel:
See @Tanych's t798-ui-tweaks branch and consider DataTables and django-tables2 plug-ins as libraries to improve functionality. This ticket also includes pagination for those tables, since in some cases it looks like it's related to the table library. 

Make sure that the column widths are optimized for the content within them. For example, the icon with the link to the platform should be a narrower width than the other columns. 

Pages affected:
* collection_set_list
* collection_set_detail
* collection_detail
* export_list
* credential_list
* monitor
 (add any I missed)

Tanych:
@lwrubel 
For `DataTables` or `django-tables2`, it depends on how much coding changes on the UI templates or views.py.
* If we take the `datatables`, most coding work will rely on the UI html templates, javascript, a little Ajax and views.py. It's kind of front-end improvement. 
* if we take the `django-tables2`, we need to make lots of change on the views.py, also the UI templates.
 We also need to add additional coding for the table.py and filter.py to implement the functions.  It depends on the [django-table2](https://github.com/bradleyayers/django-tables2) and [django-filter](https://github.com/carltongibson/django-filter).

lwrubel:
I'll leave the final decision to @justinlittman, who's evaluating the technical side of this work, but my sense is that a python/Django approach (django-tables2 and django-filter) is more in line with what we will be able to best support going forward with SFM, even if it's more work right now. I'd be interested in yours and @adityadharne's sense of the work involved with that approach, however.

justinlittman:
@adityadharne @Tanych For work on the seed table and the other tables, please make sure to coordinate your work so that the tables look, sort, page, and filter the same.

PR #876, opened 2017-07-11
Ref #853 add datatable for seeds

Tanych:
Ref #853 
* add seed info API for datatable ajax use
* add datatable for seed list
* supporting fast loading long seed list
* pagination seed list
* filtering seed list

### DIFF
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -33,4 +33,5 @@ warc==0.2.1
 djangorestframework==3.3.2
 django-filter==0.12.0
 django-datetime-widget==0.9.3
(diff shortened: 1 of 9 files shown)


justinlittman:
Let's remove "Display X records per page" since we don't offer it anywhere else. Use 25 records per page.

Is it possible to either make the table sortable? Or to be able to filter by seeds with messages?

Tanych:
@justinlittman to make it sortable, we need to define different table for different types, it would make the code more complicated. I will try to find a easy way to do so. For the message filter, I am find out whether it support the information not in the db level.

Issue #804, opened 2017-05-01
improve performance of collection_detail loading


lwrubel:
There is a lot happening on this page. @Tanych has identified some areas for improving this. It's especially evident in loading collections with a lot of seeds (e.g. Program on Extremism).

Tanych:
@lwrubel  I have done some work on the PR https://github.com/gwu-libraries/sfm-ui/pull/806. 
Other work will double-check later with the performance.

justinlittman:
Working with large collections such as the news outlets extremely slow. Fixing this should be a top priority.

lwrubel:
refs #849 and ux-cleanup work

Tanych:
@lwrubel @justinlittman 
as tested, the major cost would on getting the change log called by `diff_historical_object`. It usually takes about 80% percent of the page loading time which usually take near 2 seconds. It linearly increases by the seed size. The collection page only needs top-3 of the historical record. Need to rewrite some part of this code.

Tanych:
@justinlittman 
### Profiling the Change log
The change log is one the major factor to slow down the collection detail page, let's take a deep look at how much time cost for the change log.
## Scenario
* One Collection 
* 5000 seeds, each seed only do the add.
* Total cost nearly `11s`.

## Profiling the diff_collection_and_seeds_history
![profile_collection_and_seed](https://user-images.githubusercontent.com/3956411/27872126-baba3086-6175-11e7-8351-ea3256c4f09f.png)

The major cost is on the **append seed history** record, let's go down the function.

## Profiling the diff_object_history
![profile_object_history](https://user-images.githubusercontent.com/3956411/27872182-e82d59da-6175-11e7-92c2-b8a5d550833e.png)
The major costs are on the `list(obj.history.all)` and `generate the diff object`. For check the obj history, not sure whether there is any way to speed up. For generate the diff object, let's go future.

## Profiling the diff_historical_object
![diff_historical](https://user-images.githubusercontent.com/3956411/27872264-3332fe08-6176-11e7-84a0-f22342265600.png)
The major cost is on `diff.user = changed_historical_object.history_user`, it takes over 90% percent of the time. To get the history user, we need to find another way to speed up.

Tanych:
@justinlittman 
[...]

PR #877, opened 2017-07-11
refs #804. Improves performance of change log.

justinlittman:


Tanych:
@justinlittman 
I have tested the change log diffs for `collection_set`, `collection`, `Seeds` and `credentials`.  It works well. Also, I have reviewed the code changes, it's an efficient way to solve this problem. When we merge the changes for the seed table, we can improve the loading time a lot, from near 10 seconds to nearly `900ms` with loading about 6000 seeds.
It's good to go for the PR.

PR #879, opened 2017-07-13
Fixes #781 - Limits filter track to 400 characters

### LINKED ISSUE #781: Error when adding twitter filter seeds
When I try to add a large amount of seeds into the Track box of a Twitter Filter collection, more than about 600, and click save button, a server error occurs as the picture below.

image

image

Please note that Twitter filter API has its own limit to the number of seeds, around 400 keywords  according to Laura's search from the [Twitter Developers Forum](https://twittercommunity.com/t/streaming-api-filter-maximum-follow-name-and-how-much-data/7835) ("we can track up to 5,000 users and 400 keywords"), but I found that on the SFM UI I could add up to 600 keywords into the track box without any problem, although I got an error as below immediately after running the collection with 600 keywords. This should be considered when fixing the error, and I think a warning message preventing users from inputting more than 400 keywords should be added.
image

adityadharne:


justinlittman:
This was implemented as 400 characters, when should be 400 keywords.

Issue #895, opened 2017-09-28
Seed search on Request Export should not be case-sensitive


kerchner:
The seed search control on the Request Export page is case-sensitive, but should not be case-sensitive.

As an example, Was and was should return the same results, but return mutually exclusive results:

![image](https://user-images.githubusercontent.com/3451175/30990251-8f224666-a46e-11e7-8d02-4907182654cd.png)

![image](https://user-images.githubusercontent.com/3451175/30990275-a82579f8-a46e-11e7-8e57-15e119307954.png)

PR #898, opened 2017-10-05
refs #895. Makes export seed selector case insensitive for filtering.…

justinlittman:
… Also, removes select all option since redundant.


lwrubel:
OK to merge.

PR #900, opened 2017-10-11
T894 sharing collections

justinlittman:


lwrubel:
Suggest changing "Visibility" to "Sharing". Options are:
* Group only
* All other users

Change last sentence of help text to: Select "All other users" to share with all Social Feed Manager users.

lwrubel:
OK to squash and merge.

Issue #960, opened 2018-12-13
Support AWS Elastic Load Balancer


justinlittman:
This is based on our experience deploying SFM at Stanford Libraries.

To have AWS manage certificates, SFM must be deployed behind an Elastic Load Balancer. (The ELB acts as an HTTPS proxy, just as Nginx does within the SFM docker environment.)

An ELB performs regular health checks. However, those health checks fail with:
```
ui_1_d69f931286be               | [Thu Dec 13 10:46:19.990901 2018] [wsgi:error] [pid 68:tid 139935178061568] [remote 10.0.2.228:52162] ERROR 2018-12-13 10:46:19,990 68 django.security.DisallowedHost Invalid HTTP_HOST header: '10.0.0.120'. You may need to add '10.0.0.120' to ALLOWED_HOSTS.
```

Thus, ALLOWED_HOSTS has to be modified to permit these health check requests.

kerchner:
Closed in #961

PR #961, opened 2018-12-13
refs #960. Adds support for adding ELB to ALLOWED_HOSTS.

justinlittman:


kerchner:
@justinlittman How might we be able to test this PR?   Would we deploy to AWS using https://sfm.readthedocs.io/en/latest/install.html?highlight=AWS#amazon-ec2-installation and then set up ELB?

justinlittman:
Yes, but I'd also suggest that code review would be good enough.

PR #965, opened 2018-12-17
Upgraded versions of django and requests

kerchner:


kerchner:
Included with this PR are branches (all called `t995-dependencies-2.1` on:
sfm-ui
sfm-utils
sfm-twitter-harvester
sfm-tumblr-harvester
sfm-weibo-harvester
sfm-flickr-harvester
warcprox

kerchner:
@lwrubel note that I edited the above comment to remove warcprox

PR #969, opened 2018-12-18
Fixes #957. Correcting export segment size options

### LINKED ISSUE #957: Export segment size
As noted in #956 by @dchud: "I noticed on my collection export page (/ui/exports/EXPORT_ID/) it reported an export segment size of 1,000,000, but I had requested 100,000. The export files themselves were correct, at exactly 100,000 each, so perhaps there's a hard-coded message in there somewhere."

lwrubel:
Don't get thrown by the "100" export option that you may see in the export dropdown, depending on the value for HUNDRED_ITEM_SEGMENT param in your .env


lwrubel:
The migration file has a problem because I had HUNDRED_ITEM_SEGMENT set to True when I ran makemigrations. Fixing this branch and PR.

lwrubel:
@kerchner This is OK to review now.

kerchner:
Verified by inspecting the `<select>` tag option values in this branch vs. sfm prod.

```<option value="1000000">1,000,000</option>``` vs.  ```<option value="100000">1,000,000</option>```

Migration worked when upgrading an existing instance.  I removed the sfm_ui_1 docker container.  Upon recreation, it seemed to run the DB migration smoothly.

PR #970, opened 2018-12-19
Fixes #956. Visibility of harvesters and exporters in monitor view.

### LINKED ISSUE #956: /ui/monitor (Harvester Status) shows many harvesters
Hi, I am an adjunct lecturer at GWU and am using SFM to collect data to support an upcoming lecture. I was keeping an eye on my twitter filter and visited `/ui/monitor` to see what it had to say, and was surprised to find it showed a list of 30-40 harvesters running along with the harvest names.  Most of these names are not very revealing (e.g. a hurricane name) but perhaps some of them are.  It might be best to restrict the monitor view somewhat more.

One other thing... and sorry not to write a separate ticket for this, but I noticed on my collection export  page (`/ui/exports/EXPORT_ID/`) it reported an export segment size of 1,000,000, but I had requested 100,000.  The export files themselves were correct, at exactly 100,000 each, so perhaps there's a hard-coded message in there somewhere.

Thank you for providing this service! It's great to have this available, and I will be telling our students about it in case they are looking to study social media data.

lwrubel:
Test by comparing visibility of various harvesters and exporters owned by different users, groups, and superuser.

### DIFF
--- a/sfm/ui/views.py
+++ b/sfm/ui/views.py
@@ -1123,7 +1123,20 @@ class MonitorView(LoginRequiredMixin, TemplateView):
 
     def get_context_data(self, **kwargs):
         context = super(MonitorView, self).get_context_data(**kwargs)
-        context['harvests'] = monitor_harvests()
-        context['exports'] = monitor_exports()
+        allharvests = monitor_harvests()
+        harvests = [harvest for harvest in allharvests if
+                    has_collection_set_based_permission(harvest,
+                                                        self.request.user,
+                                                        allow_superuser=True,
+                                                        allow_staff=True,
+                                                        allow_collection_visibility=True)]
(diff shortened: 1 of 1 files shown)


kerchner:
Results as expected:
- Monitor / Harvest Status
   - User cannot see harvests from a collection in which user is not in the collection set's group (and the collection is set to group sharing)
   - User can see harvests from a collection in which user is in the collection set's group, or where collections are set to all-user visibility
   - Admin user can see all harvests
- Monitor / Exporter Status
   - User cannot see status of exports from a collection in which user is not in the collection set's group
   - User can see exports from a collection in which user is in the collection set's group, or where collections are set to all-user visibility
   - Admin user can see all export statuses

PR #979, opened 2019-04-11
updated import to reflect current version of twarc

edsu:
Someone created a twarc bug report because they were using these docs that were slightly out of date.

https://github.com/DocNow/twarc/issues/275

I adjusted the imports so the example will work. It is possible to create your credentials in a notebook if you want with:

```python
import twarc

t = Twarc(validate_keys=False)
t.configure()
```

This will prompt you for the keys inside the notebook. But I left the example as it was with the credentials already having been stored.


kerchner:
Thanks @edsu .  Regarding the second piece of this, It looks like `Twarc()` now takes care of loading the default configs (perhaps it didn't in the past?).  We appreciate the PR!

PR #990, opened 2019-07-25
fixes #905. Seed select box only shows for Selected Seeds export

### LINKED ISSUE #905: Export page selecting seeds should change option
On the export page, you can select seeds from the dropdown selector and "All seeds" remains selected as the export type. 

Either:
* Disable the dropdown selector until "Selected seeds only" is chosen as an option, or
* Change the option to "Selected seeds" if selections are made from the dropdown.

lwrubel:
Test by going to the export page for a Twitter user timeline collection with more than one seed, and choose the Selected Seeds option.

### DIFF
--- a/sfm/ui/static/ui/css/main.css
+++ b/sfm/ui/static/ui/css/main.css
@@ -118,6 +118,7 @@ body {
 
 .longseed{
     word-wrap: break-word;
+    display: none;
 }
 
 table {
--- a/sfm/ui/static/ui/js/seed_select_option.js
+++ b/sfm/ui/static/ui/js/seed_select_option.js
@@ -1,0 +1,17 @@
+$(document).ready(function() {
+        // seed multiselect available only with "Selected seeds"
+        $("input#id_id_seed_choice_0_3").on("click", function () {
+          $(".longseed").show();
+        });
+	$("input#id_id_seed_choice_0_2").on("click", function() {
+          $(".longseed").hide();
+	});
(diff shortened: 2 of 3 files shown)


kerchner:
Did not behave as expected in the following case:
1. Go to Export page.  Note that "All seeds" radio button is selected, and "None selected" control is available (non-hidden).
2. Open the "None selected" seed selector control and check 1 or more seeds.
3. Note that the "All seeds" radio button is still selected.   Once I start selecting seeds, I would expect the radio button to switch to "Selected seeds".

Note the following path as well:
1. Go to Export page.  Note that "All seeds" radio button is selected, and "None selected" control is available (non-hidden).
2. Select "Selected seeds" radio button.
3. Select "All seeds" radio button.  Note that "None selected" control is hidden.   I would suggest that #3 is okay, but the group of controls should behave consistently, so maybe it should be hidden when you first open the page as well (or never hide it in either case).

And the following path:
1. Go to Export page.  Note that "All seeds" radio button is selected. Select "Selected seeds" but do not select any seeds.
2. Press the blue Export button.  As expected, the seed selection control warns that "At least one seed must be selected"
image
3. Select a seed.  Note that the red warning is not cleared.

kerchner:
Update to previous comment:  using a fresh browser confirmed that the first two issues are not actually occurring.  Only the last issue (red warning) is still occurring.

lwrubel:
Will plan to squash commits once you've had a chance to review.

PR #994, opened 2019-07-31
Adds language parameter to twitter filter collection. Fixes #943

### LINKED ISSUE #943: Add Twitter Filter request parameter for language
As per the Twitter documentation, there are other request parameters available for Filter and Search streaming API methods:

- https://developer.twitter.com/en/docs/tweets/filter-realtime/api-reference/post-statuses-filter.html
See `language` and `filter_level` which are available at any API access level.

- https://developer.twitter.com/en/docs/tweets/filter-realtime/api-reference/post-statuses-filter.html
See  `lang`, `locale`, `geocode`, `result_type`

The use case here is a researcher who wishes to collect Tweets only in a certain langauge.

kerchner:
Also requires sfm-twitter-harvester branch https://github.com/gwu-libraries/sfm-twitter-harvester/tree/t943-twitter-lang 

Testing
- Create a twitter filter with a language parameter with one or more languages, e.g. `es,fr`.  Once the filter has collected some tweets, do an export and verify that the tweets are in the specified languages.

### DIFF
--- a/docs/collections.rst
+++ b/docs/collections.rst
@@ -112,14 +112,14 @@ Twitter filter
 ---------------
 
 Twitter Filter collections harvest a live selection of public tweets from
-criteria matching keywords, locations, or users, based on the
+criteria matching keywords, locations, languages, or users, based on the
 `Twitter filter streaming API
 <https://developer.twitter.com/en/docs/tweets/filter-realtime/overview/statuses-filter>`_. Because
 tweets are collected live, tweets from the past are not included. (Use a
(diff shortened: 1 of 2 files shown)


lwrubel:
Language parameter is not "sticking" for seed edit page. To replicate: 
1. Create a seed and add a language parameter. Save seed. 
2. On collection page, click on link to seed.
3. On seed page, click on Edit button. See that the form does not contain your previously entered text in the language parameter field.

kerchner:
The above issue has been repaired and re-squashed

Issue #250, opened 2016-05-11
Yahoo sending mail fails with SMTPSenderRefused


Tanych:
### Sign up Info
- Name: Victor
- Email: tanych5233@hotmail.com
### SFM App Info
- SFM_EMAIL_USER: tanych5233@yahoo.com
- SFM_SMTP_HOST: smtp.mail.yahoo.com
### Error Info

``` python
(501, 'Syntax error in arguments', u'webmaster@localhost')
Request Method: POST
Request URL:    http://localhost:8080/accounts/signup/
Django Version: 1.8.8
Exception Type: SMTPSenderRefused
Exception Value:    
(501, 'Syntax error in arguments', u'webmaster@localhost')
Exception Location: /usr/lib/python2.7/smtplib.py in sendmail, line 731
Python Executable:  /usr/bin/python
Python Version: 2.7.6
Python Path:    
['/opt/sfm-ui/sfm',
 '/usr/lib/python2.7',
 '/usr/lib/python2.7/plat-x86_64-linux-gnu',
 '/usr/lib/python2.7/lib-tk',
 '/usr/lib/python2.7/lib-old',
 '/usr/lib/python2.7/lib-dynload',
 '/usr/local/lib/python2.7/dist-packages',
 '/usr/lib/python2.7/dist-packages']
Server time:    Wed, 11 May 2016 16:08:35 -0500
```
### SFM Logs

``` python
sfmdevapp_1         | [Wed May 11 16:08:35.890025 2016] [:error] [pid 217:tid 140366736254720] ERROR 2016-05-11 16:08:35,888 217 django.request Internal Server Error: /accounts/signup/
sfmdevapp_1         | [Wed May 11 16:08:35.890028 2016] [:error] [pid 217:tid 140366736254720] Traceback (most recent call last):
sfmdevapp_1         | [Wed May 11 16:08:35.890029 2016] [:error] [pid 217:tid 140366736254720]   File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 132, in get_response
sfmdevapp_1         | [Wed May 11 16:08:35.890030 2016] [:error] [pid 217:tid 140366736254720]     response = wrapped_callback(request, *callback_args, **callback_kwargs)
sfmdevapp_1         | [Wed May 11 16:08:35.890031 2016] [:error] [pid 217:tid 140366736254720]   File "/usr/local/lib/python2.7/dist-packages/django/views/generic/base.py", line 71, in view
[...]

PR #996, opened 2019-08-19
Use EMAIL_FROM instead of webmaster@localhost

pratheekrebala:
Resolves https://github.com/gwu-libraries/sfm-ui/issues/250

We were trying to use Amazon SES which requires that the sender email is verified. Despite setting`SFM_EMAIL_FROM` to a verified domain, SFM would send emails using the [Django default:](https://docs.djangoproject.com/en/2.2/ref/settings/#std:setting-DEFAULT_FROM_EMAIL) `DEFAULT_FROM_EMAIL`. This was giving us the following error as SFM was trying to send email using `webmaster@localhost` instead.

`Email address is not verified. The following identities failed the check in region [REGION]: webmaster@localhost`

Updating the `DEFAULT_FROM_EMAIL` to use the value of `EMAIL_FROM` resolved the issue for us.


lwrubel:
Thanks, @pratheekrebala! Will take a look and let you know if we have questions.

Issue #993, opened 2019-07-30
Django and related dependency upgrades


kerchner:
Followon to #973:
- in sfm-ui:
    - django (currentl 2.0.13, latest is 2.3.0)
         - This will require upgrading to Postgresql 9.5
    - django-filter (currently 1.1.0, latest release is 2.2.0)

Remove datetime widget.  This will be added back in #1011

kerchner:
Ref https://docs.google.com/document/d/1uhoZKgOgSTrNE-6HiKgA50NIiwdyhPm9Iu5dWcFErR4/edit?usp=drive_web&ouid=110561461825583468817

lwrubel:
Focus on Django and related dependencies.

PR #1014, opened 2020-04-03
T993 django2 2 upgrade

lwrubel:
Upgrades from Django 2.0 to Django 2.2 and upgrades Django dependencies. Also strips out current datetimewidget which is unsupported.

Be sure to test functionality that uses sfm-ui's api, such as exports with seed selections.

This is a fix to #993


dolsysmith:
The sfm_ui_1 container has both `psycopg2==2.8.4` and `psycopg2-binary==2.8.3` installed, but `common.txt` in `./sfm-ui/requirements` references only the first. This appears to cause an error when `pip` installing the requirements for development. I commented out `psycopg2==2.8.4` and added `psycopg2-binary==2.8.3` to `common.txt`, and then the requirements installed without error, but I'm not sure that was the right thing to do.

lwrubel:
I've updated the requirements in line with what's in master, which is psycopg2-binary, using its more recent version.

PR #1017, opened 2020-04-16
Fixes #987. Upgraded to Bootstrap 4 and django-crispy-forms 1.9.0; replaced bootstrap-multiselect with bootstrap-select

### LINKED ISSUE #987: Upgrade bootstrap version
Ref:

* https://github.com/gwu-libraries/sfm-ui/blob/master/sfm/ui/templates/base.html
* https://getbootstrap.com/docs/4.3/migration/?origin_team=T08AFN77V

dolsysmith:


lwrubel:
This is looking good to me, @dolsysmith. Just a couple of small things:

1. Export detail page: Details dropdown needs the caret added. 
2. Do you still need the console.log statement in sfm/ui/static/ui/js/seed_table.js?
3. base.html still has references to multiselect.js. They're commented out, but would be good to completely remove since the replacement is working. 

BTW, I noted that we're using jquery 2.1.3 and Bootstrap 4 [can use up through 3.x ](https://github.com/twbs/bootstrap/blob/v4.0.0/package.json)(current version is 3.4.1) so we may want to look at upgrading that in the future.

PR #1021, opened 2020-04-21
T1011 datetime picker

lwrubel:
To test:
1. Do an export and select item start / item end dates (and times).
2. Do an export and select harvest start / harvest end dates (and times).
3. Create a collection and set an end date on the collecting. 

The dates should be reflected in the export.

### DIFF
--- a/sfm/ui/forms.py
+++ b/sfm/ui/forms.py
@@ -981,12 +981,12 @@ class ExportForm(forms.ModelForm):
                     HTML("""<p class="help-block">The timezone for dates entered here are {}. Adjustments will be
                     made to match the time zone of the items. For example, dates in
                     tweets are UTC.</p>""".format(settings.TIME_ZONE)),
-                    css_class="card panel-default card-body mb-3"),
+                    css_class="card panel-default card-body mb-3 input-group"),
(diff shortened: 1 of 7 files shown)


dolsysmith:
Export with item and harvest start/end date and time working as expected. There is a bit of a bug, however: if you enter a date/time in one of the fields and then decide that you actually want to leave that field blank, you can delete the date/time value. In fact, when I accidentally clicked in the harvest start date field, it inserted a date and time, and I was unable to remove it (except by cancelling the whole export and starting over).

lwrubel:
Good catch! It looks like flatpickr has a "clear" option to add onto a field. I will look and see how that might work with django-crispy-forms in these fields.

lwrubel:
@dolsysmith In addition to making the date fields clearable, I've set the default time to be 00:00 instead of 12:00, which seems like it would better align with what people are expecting to export.

lwrubel:
Looks good. I was able edit and delete dates and times. After picking a date and time, I then deleted the time, and 00:00 was restored when a datetime object was created. Tested in the export and collection edit forms. I'll squash and merge.

PR #1029, opened 2020-04-27
T1025 accessibility home

lwrubel:
To see the relevant pages, go to the login page and the home screen. 

Also sign up as a new user and reset that account's password (or at least go to the page for that).


dolsysmith:
Everything looks good.

PR #1032, opened 2020-04-28
Fixes #1023; several accessibility improvements to base.html

### LINKED ISSUE #1023: Accessibility updates for base.html
1. Change sfm-container div to a main tag. 
2. Change alerts to have a role of "status" not "alert" when yellow or green. 
3. Add skip navigation: hidden but focusable skip link at the top of the DOM that skips over the nav and breadcrumbs to the h1. 

Lower priority, but review as an option:
1. User profile dropdown in nav bar should be <button> instead of <a> with role=button

For further accessibility details see #1004

kerchner:
To test:

### Change sfm-container div to a main tag

Observe that main content is now in a `<main>` tag (previously a `<div>`)

### Change alerts to have a role of "status" not "alert" when yellow or green.

There seem to be few alerts that pass through the alert display code in `base.html` (rather than the alert display code within the body of other pages).  You can log out and get a green status alert that you have logged out, and you can log back in and get a green status alert that you logged back in.  In these cases, you can inspect the alert element and observe that it has a `role="status"` property.   If you know of some way to get an `error` alert on the main page, you can inspect that to confirm that the red alert box has `role="alert"`.

### Add skip navigation: hidden but focusable skip link at the top of the DOM that skips over the nav and breadcrumbs to the h1
_Note: I implemented this as to skip to the `<main>` block, which seems a more dependable choice than to the h1, as currently there is no h1 tag._

1.  Clear your browser's cookies OR open an incognito window.  When you first browse to the app and press <kbd>Tab</kbd>, focus should be in the cookie message text if there are links in the message text; the next focus should be the consent button.   Following this, hitting <kbd>Tab</kbd> again should unhide a `Skip to content` link in the upper left.  Pressing Enter on this link should result in the next focus being in the main body of the page - i.e. skipping the navigation bar.  The `Skip to content` link should then become hidden again.

2. Having already accepted the cookie consent in your browser, refresh the page. The first tab press on the page should un-hide the `Skip to content` link described above.

### User profile dropdown in nav bar should be instead of with role=button

Verify that the user profile dropdown in the upper right is implemented as a `<button>` tag, and that it has all the correct behaviors (the caret twists it open, the links work, etc.)

### DIFF
--- a/sfm/ui/static/ui/css/cookie_consent.css
+++ b/sfm/ui/static/ui/css/cookie_consent.css
@@ -16,13 +16,15 @@
     transform: translateY(100%);
     transition: all 500ms ease-out;
     color: #ecf0f1;
-    background: #004065
+    background: #004065;
+    visibility: hidden
 }
 
 .cookiealert.show {
     opacity: 1;
     transform: translateY(0%);
     transition-delay: 1000ms;
+    visibility: visible
 }
 
 .cookiealert a {
--- a/sfm/ui/static/ui/css/main.css
+++ b/sfm/ui/static/ui/css/main.css
@@ -245,3 +245,18 @@ table {
 .card-title {
     margin-top: 8px;
 }
+
+#skiptocontent a, #skiptocontent a:hover, #skiptocontent a:visited {
+	position:absolute;
+	left:0px;
+	top:-500px;
+	width:1px;
+	height:1px;
+	overflow:hidden;
+}
+
+#skiptocontent a:active, #skiptocontent a:focus {
+	position:static;
+	width:auto;
+	height:auto;
+}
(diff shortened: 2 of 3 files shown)


lwrubel:
The skip navigation is nice to have now and works as expected. The order of focus when the cookie consent popup is hidden is also working. User profile "button" looks good and works correctly. 

I don't have an example of an alert that should be an error on the home page, so am examining the code. If the message is error, shouldn't the class be "alert-danger"? If so, this issue pre-exists this code, but would be easier to fix with adding logic as you're doing here.

kerchner:
You're right, I'm realizing there is a mismatch between messages that this part of `base.html` is expecting, of the `django.contrib.messages` class, where `message.info` would be from the set `{debug, info, success, warning, error}` -- and Bootstrap's  `alert-danger` css class.  I think we just never had any of the error/danger type pass through this code - all of those, from what I was able to tell, pass through code in various other HTML pages.

With that said, I've updated `base.html` to handle that case better, because I wasn't able to rule out that there is or would be such an example.  (I hope the code also reads better now.)

Let me know if this looks good to you and I'll squash and merge.

lwrubel:
Looks good as updated, @kerchner.

PR #1033, opened 2020-04-28
Ticket 1004 details added in docs/limitations.rst

somanath304:
Creating a pull request for review


lwrubel:
I was able to build the docs and review the updated page. Just a few changes, @somanath304: 

- The change to .gitignore was for your local environment and shouldn't get committed to the code repo. Would you remove that change from this branch? As an example of one way to do that, see [Unmodifying a Modified File](https://git-scm.com/book/en/v2/Git-Basics-Undoing-Things).
- I thought from what I saw in the blog posts you shared that there are some earlier versions of Windows 10 that also have this Notepad problem. Dan and I have done workshops in classrooms with Windows 10 and have seen this issue. I think I saw in one of the blog posts some information about which update to Windows 10 introduces the fix. Could you add that?
- As an application, Notepad should be capitalized.

lwrubel:
I can see that the .gitgnore change is gone, thanks.  A few small typos remain:
* Notepad not NOTEPAD and WordPad no WORDPAD (sorry, my instructions were unclear earlier)
* Make sure there's a space before the parentheses, this happens twice. 
* There's a change in the file which is added characters that are not quotation marks. These aren't part of the text you were adding, maybe something accidentally added when opening in your editor? You can see this in the "Files changed" view in GitHub on this PR and this is how it's rendering:
Screen Shot 2020-04-29 at 1 49 00 PM

lwrubel:
Looks good, @somanath304.

Issue #1003, opened 2019-10-29
GW footer


kerchner:
w/accessibility links etc.

kerchner:
Do this after #987

kerchner:
@rdelaloy confirmed that GW footer is _preferable_, but accessibility text and link is _required_.  I opted to proceed with a full GW footer.

PR #1035, opened 2020-04-28
T1034 accessibility color

lwrubel:
Review that links are underlined by default, with exceptions for navigation and buttons. The button color on hover/active should now have acceptable contrast with the white text on it. I've added unique id properties to the various links to add a history note. 

I left the footer links underlined, since that will be superseded in the application of the footer update in #1003.


dolsysmith:
Everything looks good!

PR #1046, opened 2020-11-27
Docs about moving sfm-data to separate volume: Postgres and RabbitMQ user ID is 999

sebastian-nagel:
In case, sfm data is moved out from the container to a linked volume and sfm was already in use:
the user ID of PostgreSQL and RabbitMQ users is 999 (defined in the Docker containers/images). The ownership of the subfolders needs to be set accordingly. Otherwise the "db" and "mq" services will not work with the linked data volume.

### DIFF
--- a/docs/storage.rst
+++ b/docs/storage.rst
@@ -108,6 +108,9 @@ These instructions are for Ubuntu. They may need to be adjusted for other operat
 
         sudo chown -R 990:990 /sfm-data/*
         
+        sudo chown -R 999:999 /sfm-data/postgresql/
+        sudo chown -R 999:999 /sfm-data/rabbitmq/
+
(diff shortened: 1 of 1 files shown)


lwrubel:
Thanks for this recommendation for the installation docs. I'm sorry for the delay in responding. Our team really appreciates hearing what adjustments would improve the docs. We're planning to put in some work on SFM in late February and will include this in that sprint. If you've discovered anything further in the meantime about permissions set-up, let us know.

sebastian-nagel:
Thanks! We have an instance of SFM running for the project [Mobilizing Inequalities](https://www.exc.uni-konstanz.de/en/inequality/research/projects/mobilizing-inequalities/) - works nicely!

PR #1050, opened 2021-01-23
Bugfix/external services

SvenLieber:
When using a remote database in a test setup of SFM we ran into issues when using the variable `POSTGRES_HOST`.
Although it was set, an error message indicated that the database is not available.

After digging a bit I noticed that the error message `db:5432 not available after wait.` comes from the [appdeps.py](https://github.com/gwu-libraries/appdeps/blob/master/appdeps.py#L79) script of another repo.

It seems that despite the use of an environment variable which is initialized in `settings/common.py` with a default value of `db` if not set (the docker-compose hostname of Postgres) some scripts still use the hard coded `db`, and similarly `mq` for  RabbitMQ.

This pull request proposes the following changes
-  existing environment variables (with prefix `SFM_`) are used for postgres and rabbitmq to fix the hard coded issue (host, port, db name etc)
- variables used in Python (`settings/common.py` and `settings/docker_settings.py`) which are initialized with environment variables will get a default value


lwrubel:
Thanks, Sven for these suggestions. We agree it would be better not to have these settings hard-coded, and to make them configurable. We've added this PR to our next milestone.

SvenLieber:
Thank you!

One small security remark:

In case that a remote PostgreSQL DB is used one would like to encrypt the connection.
The straightforward solution we applied was outside SFM as our remote DB is behind a firewall
and we have to use a SSH tunnel/port forwarding anyway.

We establish a SSH tunnel *from the docker network* to the remote DB server which can be used
by the containers by configuring `SFM_POSTGRES_HOST` with the docker network gateway, usually `172.17.0.1` and the port accordingly  `SFM_POSTGRES_PORT` for example `1111`

```
autossh -NL 172.17.0.1:1111:localhost:5432 user@remote
```

kerchner:
@SvenLieber It looks like this PR would depend on updating an environment variable (for example, as done in https://github.com/gwu-libraries/sfm-docker/pull/28 in `example.env`) from `POSTGRES_PASSWORD` to `SFM_POSTGRES_PASSWORD`.   Would you concur that it makes sense to us to first merge #1055 and https://github.com/gwu-libraries/sfm-docker/pull/28/ _before_ this PR?

kerchner:
Noting that the environment variable names will need to be updated in docs/install.rst as well (we can do this as a separate PR if needed)

SvenLieber:
@kerchner yes I agree. Speaking about environment variables, I also updated https://github.com/gwu-libraries/sfm-docker/pull/28 as I forgot to update Postgres and RabbitMQ-related environment variables in `example.docker-compose.yml`.

Please let me know if anything else is needed from my side regarding the docs or similar.

kerchner:
This will be merged together with:
- https://github.com/gwu-libraries/sfm-docker/pull/29
- https://github.com/gwu-libraries/sfm-twitter-harvester/pull/51
- https://github.com/gwu-libraries/sfm-tumblr-harvester/pull/18
- https://github.com/gwu-libraries/sfm-flickr-harvester/pull/22
- https://github.com/gwu-libraries/sfm-weibo-harvester/pull/36

PR #1061, opened 2021-04-06
Fixes #1052: column heading bug

### LINKED ISSUE #1052: Unclarities regarding seed activations
I added a Twitter search collection with three seeds. The UI presents me the error message `Deactivate all seeds except 1 before harvesting can be turned on.`

In particular I am wondering what this means, can I only have one seed?
I also didn't find a way to deactivate a certain seed.

The Screenshot below shows the collection with all three seeds. In this example we used three seeds because because according to the documentation more than 10 search terms per query are not recommended.

image

## How can seeds be deactivated?

I also cannot find a way to activate/deactivate seeds.
In the Screenshot below I see that `is_active` was changed from `blank` to `True`, I assume together with the creation of the seed. But where can I deactivate it again? I don't see a button. The delete button is also not clickable.

image

dolsysmith:
1. Using local code in `sfm-ui`, ensure that columns display correctly for all types of seeds/collections. 
2. Confirm that warning displays appropriately with multiple active seeds for Twitter search collection.
3. Compile and review docs (under `Collection Types`: `Twitter search`).

### DIFF
--- a/docs/collections.rst
+++ b/docs/collections.rst
@@ -82,6 +82,8 @@ want to run the search once and then turn off the collection.
 See :ref:`guide-incremental-collecting` to decide whether or not to collect
 incrementally.
 
+Only one active seed can be used per search collection. If you need to run multiple searches in parallel, create a new collection for each search, each with a single seed.
+
 .. _guide-twitter-sample:
 
 .. _Twitter sample:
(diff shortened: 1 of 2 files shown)


lwrubel:
This looks good to me, @dsmith.

PR #1055, opened 2021-03-01
Replace sfm-data with more fine grained sfm-*-data

SvenLieber:
As discussed in issue https://github.com/gwu-libraries/sfm-ui/issues/1051 this PR adapts the internal folder structure of SFM to use the following directories instead of the single directory `/sfm-data`: `/sfm-db-data`, `/sfm-mq-data`, `/sfm-export-data`, `/sfm-containers-data`, `/sfm-collection-set-data`.

Merge conflicts will arise in conjunction with PR https://github.com/gwu-libraries/sfm-ui/pull/1050 as a single line in `invoke` scripts are changed by both PRs, namely `invoke.sh`, `invoke_consumer.sh`, and `invoke_runserver.sh`.

Solution: adapt respective lines to use both `/sfm-collection-set-data/collection-set` and variables for services and ports from https://github.com/gwu-libraries/sfm-ui/pull/1050

PR #1067, opened 2021-06-07
Implements fine-grained data directories

lwrubel:
Based on @SvenLieber's work in #1055 to use data directories on mounted filesystems. See https://github.com/gwu-libraries/sfm-docker/pull/30 for PRs for other repos which must be checked out as well to test.


lwrubel:
There's a problem with how warc paths are handled. For WARCs created before the fine-grained directories change, the path to each warc is stored in the database as `/sfm-data/collection_set/<collection set id>/<collection id>/` etc.

After the update, the path stored changes to:
`/sfm-collection-set-data/collection_set/<collection set id>/<collection id>/` etc.

This causes errors when exporting because the exporter is reading the paths from the database and not finding `/sfm-data/collection_set`.

lwrubel:
Addressed issue with legacy internal paths by adding to .env `DATA_VOLUME_FORMER_COLLECTION_SET` and
`DATA_VOLUME_FORMER_EXPORT`

docs/storage.rst
```diff
@@ -6,10 +6,13 @@
  Storage volumes
 -----------------
 
-SFM stores data on 2 volumes:
+SFM stores data on multiple volumes:
```
kerchner, line 9: Perhaps "in multiple directories, which may be set up as separate volumes"

kerchner (commented):
Looks good - one very mild suggestion, which you might opt to pass on

Issue #1068, opened 2021-06-07
Update docs on Amazon EC2 installation


lwrubel:
Current docs are referencing Ubuntu 14 and lack updated environment variables. https://sfm.readthedocs.io/en/stable/install.html#amazon-ec2-installation

Also update server sizing guidance.

lwrubel:
Closed with #1077

PR #1077, opened 2021-06-29
Updates EC2 installation recipe. Refs #1068.

lwrubel:
If you want to test the recipe, you'll need to change lines 90-91 to point to master, since a 2.4.0 release tag does not exist yet:

```
     - curl -L https://raw.githubusercontent.com/gwu-libraries/sfm-docker/master/example.prod.docker-compose.yml > docker-compose.yml
     - curl -L https://raw.githubusercontent.com/gwu-libraries/sfm-docker/master/example.env > .env
 ```

I've run this on an Ubuntu 18.04 image with an md5.xlarge instance (md4 is no longer available).

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -69,8 +69,8 @@ Notes:
 -------------------------
 To launch an Amazon EC2 instance running SFM, follow the normal procedure for launching an instance.
 In *Step 3: Configure Instance Details*, under *Advanced Details* paste the following in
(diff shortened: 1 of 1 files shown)


kerchner (approved):
This now works, provided one curls from master and uses non-prod docker-compose instead. 
 Instruction updates look good.   We should re-test after 2.4.0 release to ensure that the EC2 script works with zero changes at that point.

PR #1089, opened 2021-10-21
Fix off-by-one error in join_stats

sebastian-nagel:
in order to join the first two elements using the given separator.

If there are two or more item types to be shown in the "Stats" column (monitor, collection and harvest details), the first two appear glued together without any separator, eg. "1 image1 post".


lwrubel:
Thanks for this, @sebastian-nagel. Which collection types are creating the situation where this appears? I can replicate what you've reported and the change looks good. I had to replicate by manually generating multiple harvest stats objects for a harvest, so I want to make sure this isn't an indicator of another issue with harvests stats.

PR #1119, opened 2022-08-09
Added command for extracting users and collections

dolsysmith:
Django command creates a CSV of users and collections, useful to contacting users, etc.

To test, run the following in the `sfm-docker` directory (where output.csv is the name of the desired output file):

`docker-compose run ui /opt/sfm-ui/sfm/manage.py getusercollections /sfm-data-shared/output.csv`

### DIFF
--- a/sfm/ui/management/commands/getusercollections.py
+++ b/sfm/ui/management/commands/getusercollections.py
@@ -1,0 +1,72 @@
+from ui.models import *
+import json
+from django.core.management.base import BaseCommand
+import csv 
+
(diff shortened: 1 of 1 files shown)


kerchner:
@dolsysmith I think it would be helpful to add the numeric collection identifier as well (and perhaps also the parent collection_set identifier)

dolsysmith:
@kerchner If you have `DATA_SHARED_DIR` mapped to `sfm-data-shared` in `docker-compose.yml`, then the CSV should be available in whatever directory (outside the container) is represented by `DATA_SHARED_DIR`. That's how it's configured on prod.

kerchner:
@dolsysmith Got it, I see it there now in the directory mapped to DATA_SHARED_DIR.

PR #1122, opened 2022-08-09
Twv2 UI datefields

adhithyakiran:
Date fields restored for twitter search version2
Added warning message for twitter academic search seed


dolsysmith:
@adhithyakiran I am getting an error message when trying to add an Academic Search seed:

Screen Shot 2022-08-11 at 9 00 39 AM

It seems to be looking for a file that doesn't exist?

dolsysmith:
@adhithyakiran It looks good, but can you add a space after the period and before `See Twitter` in the alert, and also before `Limits are approximate` in the help text?

Once you do that, I think you can squash and merge this PR. (We usually do squash and merge so that the commit history doesn't get too big.)

dolsysmith:
Hi @adhithyakiran, I tested your changes to the credentials form. Unfortunately, by changing the Python name of the form in `forms.py`, it breaks references to this element elsewhere in the code:

```
  File "/opt/sfm-utils/sfmutils/harvester.py", line 239, in on_message
    self.harvest_seeds()
  File "twitter_harvester.py", line 48, in harvest_seeds
    self._create_twarc2()
  File "twitter_harvester.py", line 81, in _create_twarc2
    self.twarc = Twarc2(self.message["credentials"]["consumer_key"],
KeyError: 'consumer_key'
```
Can you try instead keeping the element names as they were but adding a `label` attribute? See [here](https://stackoverflow.com/questions/636905/django-form-set-label) for some examples that might work.

sfm/ui/templates/base.html
```diff
@@ -9,6 +9,9 @@
     <meta name="author" content="">
     <link rel="shortcut icon" type="image/png" href="{% static 'ui/img/favicon.ico' %}" />
 
+    {% block title2 %}
```
dolsysmith, line 12: @adhithyakiran Is this block still needed? if not, we should probably delete.

adhithyakiran: I have deleted those block, we don't need that

PR #1136, opened 2023-03-13
T1104 v2 filter stream

### LINKED ISSUE #1105: Update UI for Twitter 2 search/streaming functionality
### To Do

- See Laura's commits: [https://github.com/gwu-libraries/sfm-ui/tree/twitter-v2](https://github.com/gwu-libraries/sfm-ui/tree/twitter-v2)
- Because of the tiered access to the v. 2 API, users need to see error messages/in-app documentation related to the following:
   - Using (streaming) operators they are not permitted to use
   - Trying to add more stream rules than they are allowed (not sure if this triggers an error or just replaces the rules currently in use)
   - Exceeding the monthly limit
   - Trying to use the `search_all` endpoint without Academic Research access
- Users can select which fields they want to return, including [annotations](https://developer.twitter.com/en/docs/twitter-api/annotations/overview). We need to provide a mechanism for users to do this.

#### Streaming API

- Add a way for users to review their current streaming rules (since they're limited by account type to either 5, 25, or 1,000).
- Add support for new query operators in creating stream rules.

#### Search API

- Test Laura's editions

### LINKED ISSUE #1126: Upgrade to Django 3.2+
Django 3.2 is a long-term supported release, and it is still compatible with Postgres 9 (ref #1124 ).  We will upgrade separately to Django 4.2 in order to postpone the Postgres upgrade.

dolsysmith:
Resolves #1126 
Resolves #1105 

### Contains
- Upgrade to Django 3.2 (and dependencies)
- Forms for Twitter v.2 Filtered Stream 
  - collection form
  - seed form
- Removal of Twitter v.1 Sample collection type (deprecated)

### Testing
- Verify that user can create SFM account and add credentials (both Twitter v1 and Twitter v2)
- Verify creation of Twitter v1 collections: Search, Filter, User Timeline
- Verify that the Sample collection type is unavailable
- Verify creation of Twitter v2 collections: Search (2), Academic Search, Filtered Stream, User Timeline (2)
- Verify creation of seeds for the above collection types
- For Filtered Stream collection, verify that user can add multiple streaming rules
   - `rule` field should be required 
   - `tag` field should be optional
   - Verify that user can edit rules already created

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,8 @@
 FROM gwul/sfm-base@sha256:0b80a3d3562cdb4d631fbb55b9bd24889312838cbd27cd33e14cc0c18405f007
 MAINTAINER Social Feed Manager <sfm@gwu.edu>
 
+ARG build_version=release
+
 # Install apache
 RUN apt-get update && apt-get install -y \
     apache2=2.4* \
@@ -9,7 +11,7 @@ RUN apt-get update && apt-get install -y \
 
 ADD . /opt/sfm-ui/
 WORKDIR /opt/sfm-ui
-RUN python -m pip install -r requirements/common.txt -r requirements/release.txt
+RUN python -m pip install -r requirements/common.txt -r requirements/${build_version}.txt
 
 # Adds fixtures.
 ADD docker/ui/fixtures.json /opt/sfm-setup/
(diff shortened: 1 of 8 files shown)


dolsysmith:
Typo under `Start Time` parameter on Seed form for Search 2: `ignore` should be `ignored`.

dolsysmith:
Update version number in UI footer from 2.5 to 3.0

dolsysmith:
Disable v1.1 collection types and credential.

